PR #23, opened 2023-09-23
Use `attestantio/go-eth2-client` types & RPC client implementation

pk910:
This PR migrates the internal beaconchain types to the types from [attestantio/go-eth2-client](https://github.com/attestantio/go-eth2-client).

Improvements compared to previous implementation:
* Orphaned and old unfinalized blocks are now persisted to DB in SSZ format. 
  This makes the DB a lot smaller compared to the previous JSON based approach.
* The built in response validation from go-eth2-client filters out some invalid rpc responses that previously caused various issues.

Todo:
* [x] Wait for [go-eth2-client PR #74](https://github.com/attestantio/go-eth2-client/pull/74) and get rid of the temporary getters
* [ ] Wait for [go-eth2-client PR #75](https://github.com/attestantio/go-eth2-client/pull/75) and disable initial connection check [here](https://github.com/pk910/dora-the-explorer/blob/external-eth-types/rpc/beaconapi.go#L42-L43)

PR is currently blocked on the second todo point, as this prevents the explorer from using prysm nodes.


pk910:
Merging as it blocks other commits that are based on this.

PR #72, opened 2024-07-04
feat: client peers visualisation

skylenet:
Some notable changes:
- Added [sprig](https://masterminds.github.io/sprig/) to be used on any template 
- Added peer info to CL nodes
- Made it possible to visualise the peer info on the client webpage. Graph is generated using [VisJS](https://visjs.github.io/vis-network/examples/).
- Added identicon endpoint. It's useful to generate identicons based on the peer IDs. 

Screenshot:
image


types/models/clients.go
```diff
@@ -1,20 +1,47 @@
...
+	Index       int        `json:"index"`
+	Name        string     `json:"name"`
+	Version     string     `json:"version"`
+	HeadSlot    uint64     `json:"head_slot"`
+	HeadRoot    []byte     `json:"head_root"`
+	Status      string     `json:"status"`
+	LastRefresh time.Time  `json:"refresh"`
+	LastError   string     `json:"error"`
+	PeerId      string     `json:"peer_id"`
+	Peers       []*v1.Peer `json:"peers"`
```
pk910, line 26: Please make a separate struct for `v1.Peer` with only the relevant information for the UI.
Small thing here, but UI models are cached in memory or in external cache, so they should be as small as possible :)

indexer/consensus_client.go
```diff
@@ -177,6 +189,19 @@ func (client *ConsensusClient) checkClient() error {
...
+	if err != nil {
+		return fmt.Errorf("could not get node peer id: %v", err)
+	}
+
+	// get peers
+	peers, err := client.rpcClient.GetNodePeers()
+	if err != nil {
+		return fmt.Errorf("could not get peers: %v", err)
+	}
+	client.peers = peers
```
pk910, line 203: the peer list loaded in here is not getting refreshed for ready clients as the whole `checkClient()` function is only called for offline clients to check if the client is ready to be used as data source.
once clients are ready, the execution is "locked" inside the `processClientEvents()` "process events" loop.

I think you should still keep it there to refresh the clients list for syncing clients, but also add the refresh to the process events loop.

indexer/consensus_client.go
```diff
@@ -102,6 +108,12 @@ func (client *ConsensusClient) GetLastClientError() string {
 	return client.lastClientError.Error()
 }
 
+func (client *ConsensusClient) GetNodePeers() []*v1.Peer {
+	client.cacheMutex.RLock()
+	defer client.cacheMutex.RUnlock()
+	return client.peers
+}
```
pk910, line 115: cacheMutex locking not required here :) 
but more important: 
maybe return a empty list here if a client is offline (not syncing) for some time?

pk910 (changes requested):

types/models/clients.go
skylenet: Also has some security aspect of splitting up the view models to not accidentally exposing more than whats needed 👍

indexer/consensus_client.go
skylenet: I've made the peer check happen during epoch changes as per recommendation to not query it too often.

pk910 (approved):

PR #76, opened 2024-07-11
execution layer clients page

skylenet:


indexer/execution_client.go
```diff
@@ -205,6 +239,12 @@ func (client *ExecutionClient) processClientEvents() error {
 				return err
 			}
 			client.lastHeadRefresh = time.Now()
+
+			// update node peers
+			if err = client.updateNodePeers(ctx); err != nil {
+				return fmt.Errorf("could not get node peers for %s: %v", client.clientName, err)
+			}
```
pk910, line 246: can you wrap this into a if block that makes the refresh less frequent?
currently it polls for every slot, which is a bit too often.
once per epoch, like for the CL peers, should be fine :)

rpc/executionapi.go
```diff
@@ -115,6 +116,18 @@ func (ec *ExecutionClient) GetClientVersion(ctx context.Context) (string, error)
 	return result, err
 }
 
+func (ec *ExecutionClient) GetAdminPeers(ctx context.Context) ([]*p2p.PeerInfo, error) {
+	var result []*p2p.PeerInfo
+	err := ec.rpcClient.CallContext(ctx, &result, "admin_peers")
```
pk910, line 121: that's a quite heavy namespace we require here (`admin`).
that namespace might not be available for external EL nodes, so the error produced here should not bubble up and make the whole ExecutionClient unusable.
maybe catch these errors in the `updateNodePeers` function and just set a empty peerlist/peerid if a error is thrown?

handlers/pageData.go
```diff
@@ -126,13 +127,18 @@ func createMenuItems(active string) []types.MainMenuItem {
...
-				Path:  "/clients",
+				Label: "Consensus",
+				Path:  "/clients/consensus",
 				Icon:  "fa-server",
 			},
+			{
+				Label: "Execution",
+				Path:  "/clients/execution",
+				Icon:  "fa-circle-nodes",
+			},
```
pk910, line 141: can you add this menu point dynamically if at least 1 execution layer client is configured?
dora can be run without a execution layer configured (There are some details missing, but works in general).
we should not show pages that are known to not work at all if the configuration for it is missing (similar to the MEV Relays menu entry).

pk910 (changes requested):

pk910 (approved):

Issue #49, opened 2024-01-26
Feature: Load specs from beacon nodes


pk910:
The explorer should be able to load the current network specs from the connected beacon nodes.
This will make the explorer easier to run on custom devnets.

PR #83, opened 2024-07-29
Indexer reimplementation

pk910:
Indexer re-implementation & code cleanup.

The overall goal of this PR is to re-implement the whole indexer logic.
The existing indexer logic is quite unstable in case of forks, is very complex and got unmaintainable over time.

* Decouple indexer from client pool logic
  * [x] Copy / Amend consensus client pool package from [assertoor](https://github.com/ethpandaops/assertoor/tree/master/pkg/coordinator/clients/consensus)
  * [x] Implement lightweight execution client pool (based on [assertoor](https://github.com/ethpandaops/assertoor/tree/master/pkg/coordinator/clients/execution), but without block caching)
  * [x] Update `cl_clients` page to use consensus pool as datasource
  * [x]  Update `el_clients` page to use execution pool as datasource
* Re-Implement Indexer
  * [x] Load specs from consensus pool instead of local file (#49)
  * [x] Proper support for chain forks
    * [x] Decouple block & epoch stats cache, add separate epoch stats loader
    * [x] Improve validator set caching (re-use validator struct from previous validator sets if all values match)
    * [x] Compute proposer/attester/sync duties from state (avoid loading duties from RPC as the api does not support loading duties for non canonical forks)
    * [x] Implement fork handling & canonical fork selection
  * [x] Persist unfinalized blocks in database
  * [x] Persist unfinalized epoch stats in database
  * [x] Implement block cache processing routines
    * [x] Block body pruning
    * [x] Finalized processing
    * [x] Orphaned block processing
  * [x] Implement epoch history synchronization


pk910:
fixes #49 as specs are coming from consensus pool now

PR #130, opened 2024-09-17
Add withdrawal requests & consolidation requests page

pk910:
Add a simple withdrawals requests page similar to the other filtered list views:
image


pk910:
Added consolidation requests page too:
image

skylenet (approved):
Awesome 🚀

PR #132, opened 2024-09-23
PeerDAS: Column distribution table

skylenet:
This feature is disabled by default and has to be enabled via config. `FRONTEND_SHOW_PEER_DAS_INFOS=true` or in the yaml config file via `frontend.showSensitivePeerInfos: true`. 

Add capabilities to extract information related to peerDAS from all peers. With this we can generate a column distribution table where we have an overview about columns that are under peer custody.

Some notable changes/features:
- Always showing the most up-to-date ENR info from peers based on the peer that has the highest `seq` field. ENRs from other peers can be out of sync. Example NodeA sees PeerA with ENR `seq=10` while NodeB still reports PeerB with an ENR with an older `seq=2`. 
- When clicking on a node on the PeerDAS table, you'll get a modal popup with detailed node infos. You can also click its peers and navigate from there onwards.
- Allows filtering the PeerDAS Table by peer type ( internal, external, supernode) and also using the search box. 
- Added a warning button that appears when there's some problem. Clicking on the button results on a modal appearing with more details about the problem. These can be:
  - Could not obtain PeerDAS relevant spec values from the beacon API
  - A peer doesn't have an ENR. Some nodes don't have an ENR for a given peer, thus we can't calculate the right distribution across columns due to the missing `CSC` field from the ENR. When this happens, we just default to the default custody requirement from the spec
  - The ENR exists but the `csc` field is not present. 
  - There are empty columns, without a peer in it. 

Example screenshots:

## PeerDAS column overview table
image
## Peer details modal when clicking on a node of the PeerDAS column table
image
## Example warnings
image
## Warning about an outdated ENR, and that the data is based on another node.
image


templates/clients/clients_cl.html
```diff
@@ -1,4 +1,30 @@
...
+  <!-- Peer details modal -->
+  <div class="modal modal-xl" id="peerDetailsModal" tabindex="-1" role="dialog" aria-labelledby="peerDetailsModalTitle" aria-hidden="true">
+    <div class="modal-dialog modal-dialog-centered modal-xl" role="document">
+      <div class="modal-content">
+        <div class="modal-header">
+          <h5 class="modal-title" style="overflow-wrap: anywhere;">
+            <svg id="peerDetailsModalTitleImage" class="peer-table-icon" style="width:50px;height:50px;" data-jdenticon-value="user127"></svg>
+            <span id="peerDetailsModalTitle"></span>
+          </h5>
+          <button type="button" class="close" data-dismiss="modal" aria-label="Close"  onclick="$('#peerDetailsModal').modal('hide')">
```
pk910, line 14:
```suggestion
          <button type="button" class="close btn btn-outline-secondary" data-dismiss="modal" aria-label="Close"  onclick="$('#peerDetailsModal').modal('hide')">
```
design fix

templates/clients/clients_cl.html
```diff
@@ -38,6 +64,181 @@ <h2 class="accordion-header">
...
+                </p>
+              </div>
+
+              <!--- Peer DAS Warnings modal -->
+             <div class="modal fade" id="peerDasWarningsModal" tabindex="-1" role="dialog" aria-labelledby="peerDasWarningsModal" aria-hidden="true">
+              <div class="modal-dialog modal-lg" role="document">
+                <div class="modal-content">
+                  <div class="modal-header">
+                    <h5 class="modal-title" id="peerDasWarningsModal">👀 PeerDAS: Some problems were detected</h5>
+                    <button type="button" class="close" data-dismiss="modal" aria-label="Close" onclick="$('#peerDasWarningsModal').modal('hide')">
```
pk910, line 125:
```suggestion
                    <button type="button" class="close btn btn-outline-secondary" data-dismiss="modal" aria-label="Close" onclick="$('#peerDasWarningsModal').modal('hide')">
```
design fix

static/js/common.js
```diff
@@ -0,0 +1,11 @@
+  // Fix bootstrap backdrop stacking when having multiple modals
+  $(document).on('show.bs.modal', '.modal', function() {
+    const zIndex = 1000 + Math.max(...Array.from(document.querySelectorAll('*')).map((el) => +el.style.zIndex));
```
pk910, line 3: this is very slow,  takes 140ms on my side ;)
might wanna filter more via the selector, eg. taking only modal divs into account.

static/js/common.js
```diff
@@ -0,0 +1,11 @@
+  // Fix bootstrap backdrop stacking when having multiple modals
```
pk910, line 1: might wanna move this logic into the `explorer.js`, which is loaded on all pages too and contains the other common stuff :)

static/js/common.js
skylenet: Yeah, just removed it and used a static high zindex.

templates/clients/clients_cl.html
skylenet: Good eye! Instead of your suggestion, I went with `class="btn-close"`, looks a bit better.

pk910 (approved):

PR #136, opened 2024-09-25
Add json content handler for /forks

0xTylerHolmes:
I've been using Dora for some testing and really like exposing the /forks page to allow dumping the forks as JSON. I would be eager to have this merged. Do you have any recommendations?


handlers/forks.go
```diff
@@ -31,6 +33,21 @@ func Forks(w http.ResponseWriter, r *http.Request) {
 		handlePageError(w, r, pageError)
 		return
 	}
+
+	if r.Header.Get("Content-Type") == "application/json" {
```
pk910, line 37: Should check for the `Accept` header here.
`Content-Type` in requests describes the type of the data being delivered with the request.
Its used for `POST`/`PUT`/`PATCH` requests only

0xTylerHolmes: Ah gotcha, how does this look to you?

pk910:
Heya @0xTylerHolmes, thanks for your PR.
I think it looks fine for this page :)

These page models are normally not intended to be served as json, because they contain sensitive information in other pages.
However, for the forks page it's fine, but we might change it in future if we add proper api routes.

pk910 (approved):

PR #143, opened 2024-10-09
fix `make devnet-run` for prysm nodes

pk910:


skylenet:
Instead of this change, I think it makes more sense to make prysm run on port 4000 on the ethereum-package: https://github.com/ethpandaops/ethereum-package/pull/802

PR #144, opened 2024-10-10
add execution block hash index to block cache

pk910:
This PR adds a execution block hash to block reference map to the block cache.
This map significantly lowers the processing time for the `getBlocksByExecutionBlockHash` function, which is more heavily used with #141.

The negative effect of this change is the memory overhead.
The map holds all execution hashes & references to blocks for all unfinalized blocks, 
which roughly sums up to about 2.5M memory for a week of unfinality (50k blocks).

I think the overhead is acceptable compared to iterating the full unfinalized block range for each call to `getBlocksByExecutionBlockHash`.

### DIFF
--- a/indexer/beacon/blockcache.go
+++ b/indexer/beacon/blockcache.go
@@ -10,25 +10,29 @@ import (
 	"github.com/ethpandaops/dora/db"
 )
 
+var zeroHash = phase0.Hash32{}
+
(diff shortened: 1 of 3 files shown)


Maliksb11:
Ok

skylenet (approved):

PR #167, opened 2024-11-04
Check for double deposits on mini launchpad

pk910:
Check for double deposits on mini launchpad:
image

image

### DIFF
--- a/handlers/submit_deposit.go
+++ b/handlers/submit_deposit.go
@@ -1,12 +1,17 @@
 package handlers
 
 import (
+	"encoding/json"
 	"errors"
+	"fmt"
 	"net/http"
 	"time"
 
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/sirupsen/logrus"
 
+	"github.com/ethpandaops/dora/db"
+	"github.com/ethpandaops/dora/dbtypes"
 	"github.com/ethpandaops/dora/services"
 	"github.com/ethpandaops/dora/templates"
 	"github.com/ethpandaops/dora/types/models"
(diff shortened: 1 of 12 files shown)


handlers/submit_deposit.go
```diff
@@ -82,3 +101,73 @@ func buildSubmitDepositPageData() (*models.SubmitDepositPageData, time.Duration)
...
 	return pageData, 1 * time.Hour
 }
+
+func handleSubmitDepositPageDataAjax(w http.ResponseWriter, r *http.Request) error {
+	query := r.URL.Query()
+	var pageData interface{}
+
+	switch query.Get("ajax") {
+	case "load_deposits":
+		if r.Method != http.MethodPost {
```
skylenet, line 111: Just a minor nit: In theory `/ajax=load_deposits`  should be a `GET` request cause no data is being changed.

skylenet (approved):

ui-package/src/components/SubmitDepositsForm/DepositEntry.tsx
```diff
@@ -79,6 +87,109 @@ const DepositEntry = (props: IDepositEntryProps): React.ReactElement => {
...
+                      </div>
+                    </div>
+                  </div>
+                  <div className="d-flex">
+                    <div className="tx-details-label">
+                      Block:
+                    </div>
+                    <div className="tx-details-value">
+                      <div className="d-flex">
+                        <span className="flex-grow-1 text-truncate">{tx.block}</span>
```
skylenet, line 122: Maybe a link to the block explorer and respective block number would be a nice to have.

skylenet: Nevermind. I just saw you have a link to the TX if the explorerUrl exists.

handlers/submit_deposit.go
pk910: I'm not sure about this, `GET` data is limited by url size, so it's only suitable for limited amount of information.
In this situation I'm sending in a unbound list of public keys, each consuming 96Bytes in hex format.
When using the GET approach we'll hit url size limits at some point with large deposit files.
eg. 100 pubkeys will already consume 9600 bytes, leaving out any delimiter, argument name or other stuff that'S needed for valid urls ;)

skylenet: Yep, that's a very valid reason to not use GET 👍  Makes sense to move that to the HTTP body.

PR #175, opened 2024-11-22
WIP: use client side rendering for clients list

pk910:


utils/enr.go
```diff
@@ -37,13 +37,7 @@ func GetKeyValuesFromENR(r *enr.Record) map[string]interface{} {
 	fields["seq"] = r.Seq()
 
 	// Get signature
-	n, err := enode.New(enode.ValidSchemes, r)
-	if err == nil {
-		record := n.Record()
-		if record != nil {
-			fields["signature"] = "0x" + hex.EncodeToString(record.Signature())
-		}
-	}
```
pk910, line 46: @skylenet 
Has there been a reason for parsing the enr record again to get the signature?
This consumed a significant amount of time for the clients page on the ntf devnet.

skylenet: Good catch, totally unnecessary. Just out of curiosity, I was wondering why you mention "significant amount of time", because I was expecting it to just duplicate some structs, but it seems that  enode.New() will then actually verify the signature, which can create some overhead.

skylenet (approved):
🚀

utils/enr.go
pk910: Yea, the signature check is what made it quite time consuming.
I've also added a enr map to the page handler,  so the enr's are not re-parsed over and over again, has a huge impact on page loading time :)

PR #184, opened 2024-12-02
fix: make timestamp copy btn work in epoch page

syjn99:
Hi, I found that copy button for epoch timestamp doesn't work. `copyTs` seems a legacy function, so I changed it as other pages do.


pk910:
nice catch :)
probably a copy paste leftover from beaconcha.in :D

PR #238, opened 2025-02-09
add warnings for invalid el triggered consolidations & withdrawals

### LINKED ISSUE #229: You cannot do a consolidation request from a `0x01` source validator to a different `0x01`
From what I understand, you cannot do a consolidation request from a `0x01` source validator to a different `0x01` target validator. The source and the target validator must be the same for an in-place upgrade from `0x01` to a `0x02` validator or the target validator must be a `0x02` one. This is currently possible through the dora UI and it will show in the consolidation requests tab on a validator's page.

See https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/beacon-chain.md#new-process_consolidation_request

```python
    # Verify that target has compounding withdrawal credentials
    if not has_compounding_withdrawal_credential(target_validator):
        return
```

pk910:
adds warnings for probably invalid requests:
image

there are warnings for:
* consolidation requests to a validator without 0x02 credentials
* consolidation request from a validator that is younger than the shard committee period (not consolidate-able)
* withdrawal request for a validator that is younger than the shard committee period (not withdrawable)

fixes #229, but keeps the possibility to submit invalid requests for testing purposes.
cc @remyroy


pk910:
#236 should also make the outcome of requests more clear.

skylenet (approved):

PR #235, opened 2025-02-06
very basic dora api for devnet tooling support

pk910:
This PR implements a very basic api to dora that is compatible with the [beaconcha.in api](https://beaconcha.in/api/v1/docs/index.html#/):
* [x] `/api/v1/validator/{indexOrPubkey}`
* [x] `/api/v1/validator/eth1/{eth1address}`
* [x] `/api/v1/validator/withdrawalCredentials/{withdrawalCredentialsOrEth1address}`
* [ ] `/api/v1/ethstore/latest`
* [x] `/api/v1/epoch/{epochIndex}`

more todo:
* [ ] Swagger api docs

These endpoints are required by downstream tools we'd like to run on devnets as well (eg. launchpad).
API needs to be enabled via the `api.enabled` setting in the dora config.

PR #263, opened 2025-03-05
Return correct response for ApiWithdrawalCredentialsValidators

remyroy:
Fixes issue in https://github.com/ethpandaops/dora/pull/235#issuecomment-2701867520

This is my first time working with Go and I'm not too sure what I'm doing or how to test it but this feel corect to me 😉


pk910:
thanks for fixing :)
looks good, I've clearly missed that one :)

pk910 (approved):

PR #272, opened 2025-03-11
Display ILs per slot

jihoonsong:
This PR make Dora to listen to IL events, store ILs and display them in the slot page.

Clients can emit events over the same IL so Dora performs the duplication check before storing ILs. When there is an equivocation, both ILs by the equivocator are stored and be displayed. Currently, ILs are stored on memory.

One thing to notice is it assumes no fork. This would be our future work and it should consider IL committee root to cope with multiple forks properly.

To run this successfully, you could use this [Kurtosis config](https://github.com/jihoonsong/local-devnet-focil/blob/main/kurtosis/kurtosis.config.registry.yaml). You should be able to see ILs after `eip7805` fork is activated as below:

Screenshot 2025-03-11 at 1 34 05 AM

### DIFF
--- a/clients/consensus/client.go
+++ b/clients/consensus/client.go
@@ -52,6 +52,7 @@ type Client struct {
 	blockDispatcher         Dispatcher[*v1.BlockEvent]
 	headDispatcher          Dispatcher[*v1.HeadEvent]
 	checkpointDispatcher    Dispatcher[*v1.Finality]
+	inclusionListDispatcher Dispatcher[*v1.InclusionListEvent]
 }
 
 func (pool *Pool) newPoolClient(clientIdx uint16, endpoint *ClientConfig) (*Client, error) {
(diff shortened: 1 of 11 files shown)


jihoonsong:
Hey @pk910, thanks for the review and feedback! I've addressed it with following changes:
1) Check duplication by comparing IL's signature instead of comparing transactions one by one.
2) At 1slot N`, it should fetch ILs submitted in `slot N-1`. Found and fixed this bug.
3) Display whether each IL transaction is included in a block.

Now it looks like this:
Screenshot 2025-03-13 at 12 37 40 AM

But now we encountered to a performance issue. When a user clicks the IL tab, 1,000+ transaction decoding is processed. This is quite a heavy computation, making it laggy. I can come up with some possible solutions including parallelize transaction decoding, cache decoded transaction using `map[txByte]decoded_tx`, lazy loading IL transactions, etc.

jihoonsong:
I'll be working on starting with Electra genesis and some other issues and get back to this. If you come up with any better idea in the meantime, please let me know :)

PR #284, opened 2025-03-16
implement separate block store

pk910:
This PR allows running dora with checkpoint synced nodes only.
Historic blocks are stored in a separate block database (using pebble or S3).

This is an optional feature which is not enabled by default as it causes some disk overhead / cloud costs.

This PR also adds a new executable `dora-utils`, which provides some helpful toolings for running dora:
* database migration tool (copy database between sqlite/pgsql)
* blockdb sync tool (fill blockdb without the need to resync the whole dora index)


skylenet (approved):

config/default.config.yml
```diff
@@ -104,3 +104,10 @@ database:
     user: ""
     password: ""
     name: ""
+
+# separate block db for storing block bodies (no archive beacon node required)
+blockDb:
+  engine: "none" # pebble / none
```
skylenet, line 110: Should also add `s3` to the comment and also have some commented out s3 config example?

PR #295, opened 2025-03-22
rename rocketpool validators

0xobjectObject:


pk910:
merged into #296 as there are more name changes for hoodi :)

PR #326, opened 2025-04-16
Add missing deposits API for launchpad support

remyroy:
This adds a compatible missing validator deposits api to the one offered by beaconcha.in on https://beaconcha.in/api/v1/docs#tag/validators/GET/api/v1/validator/{indexOrPubkey}/deposits to support the launchpad project.

I didn't test this much and I'm not sure how to do it properly but it seems good from a code point of view.

### DIFF
--- a/handlers/api/general.go
+++ b/handlers/api/general.go
@@ -1,15 +1,53 @@
 package api
 
 import (
+	"encoding/hex"
 	"encoding/json"
+	"fmt"
+	"math"
 	"net/http"
+	"strconv"
+	"strings"
 
+	"github.com/attestantio/go-eth2-client/spec/phase0"
+	"github.com/ethpandaops/dora/services"
 	"github.com/sirupsen/logrus"
 )
 
 type ApiResponse struct {
 	Status string      `json:"status"`
 	Data   interface{} `json:"data"`
+}
+
+func parseValidatorParamsToIndices(origParam string, limit int) (indices []phase0.ValidatorIndex, err error) {
+	params := strings.Split(origParam, ",")
+	if len(params) > limit {
+		return nil, fmt.Errorf("only a maximum of %d query parameters are allowed", limit)
+	}
+	for _, param := range params {
+		if strings.Contains(param, "0x") || len(param) == 96 {
+			pubkey, err := hex.DecodeString(strings.Replace(param, "0x", "", -1))
+			if err != nil {
+				return nil, fmt.Errorf("invalid validator-parameter")
+			}
+
+			indice, found := services.GlobalBeaconService.GetValidatorIndexByPubkey(phase0.BLSPubKey(pubkey))
+			if !found {
+				return nil, fmt.Errorf("validator pubkey %s not found", param)
+			}
+			indices = append(indices, indice)
+		} else {
+			index, err := strconv.ParseUint(param, 10, 64)
+			if err != nil {
+				return nil, fmt.Errorf("invalid validator-parameter: %v", param)
+			}
+			if index < math.MaxInt64 {
+				indices = append(indices, phase0.ValidatorIndex(index))
+			}
+		}
+	}
+
+	return
 }
 
 func sendBadRequestResponse(w http.ResponseWriter, route, message string) {
--- a/handlers/api/general.go
+++ b/handlers/api/general.go
@@ -50,6 +50,56 @@ func parseValidatorParamsToIndices(origParam string, limit int) (indices []phase
 	return
 }
 
+func parseValidatorParamsToPubkeys(origParam string, limit int) (pubkeys [][]byte, err error) {
+	params := strings.Split(origParam, ",")
+	if len(params) > limit {
+		return nil, fmt.Errorf("only a maximum of %d query parameters are allowed", limit)
+	}
+
+	indices := []phase0.ValidatorIndex{}
+	indicePubkeyPos := map[int]phase0.ValidatorIndex{}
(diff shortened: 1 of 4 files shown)


pk910:
Heya @remyroy,
Thanks for the implementation :) 
It looks very good at first glance, unfortunately I think I found a logic problem that could lead to problems especially with the new electra deposit queue.

So, let me explain my finding:
The new api endpoint accepts a list of indices or pubkeys and should return the execution layer deposit transactions for these pubkeys.
The current implementation resolves the list of pubkey/indices to a list of indices, which is then used to the lookup the pubkeys via the validator set. This might not work if the pubkey is not part of the validator set yet, which might be the case for pending deposits. So these requested pubkeys get lost during the conversion to indices and back.
That's probably not ideal for the launchpad usecase, which tries to find duplicate deposits with this api call, right?

handlers/api/general.go
```diff
@@ -12,6 +19,37 @@ type ApiResponse struct {
...
+			}
+			if index < math.MaxInt64 {
+				indices = append(indices, phase0.ValidatorIndex(index))
+			}
+		}
+	}
+
+	return
+}
+
```
pk910, line 52:
```suggestion

func parseValidatorParamsToPubkeys(origParam string, limit int) (pubkeys [][]byte, err error) {
	params := strings.Split(origParam, ",")
	if len(params) > limit {
		return nil, fmt.Errorf("only a maximum of %d query parameters are allowed", limit)
	}

	indices := []phase0.ValidatorIndex{}
	indicePubkeyPos := map[int]phase0.ValidatorIndex{}

	for _, param := range params {
		if strings.Contains(param, "0x") || len(param) == 96 {
			pubkey, err := hex.DecodeString(strings.Replace(param, "0x", "", -1))
			if err != nil {
				return nil, fmt.Errorf("invalid validator-parameter")
			}

			pubkeys = append(pubkeys, pubkey)
		} else {
			index, err := strconv.ParseUint(param, 10, 64)
			if err != nil {
				return nil, fmt.Errorf("invalid validator-parameter: %v", param)
			}
			if index < math.MaxInt64 {
				indicePubkeyPos[len(pubkeys)] = phase0.ValidatorIndex(index)
				pubkeys = append(pubkeys, []byte{}) // placeholder
				indices = append(indices, phase0.ValidatorIndex(index))
			}
		}
	}

	if len(indices) > 0 {
		// lookup pubkeys from validator set
		validators, _ := services.GlobalBeaconService.GetFilteredValidatorSet(&dbtypes.ValidatorFilter{
			Indices: indices,
		}, false)

		validatorsMap := map[phase0.ValidatorIndex]v1.Validator{}
		for _, validator := range validators {
			validatorsMap[validator.Index] = validator
		}

		for i, indice := range indicePubkeyPos {
			if validator, ok := validatorsMap[indice]; ok {
				pubkeys[i] = validator.Validator.PublicKey[:]
			}
		}
	}

	return
}
```

handlers/api/validator_deposits_v1.go
```diff
@@ -0,0 +1,121 @@
...
+
+		err := decoder.Decode(req)
+		if err != nil {
+			sendBadRequestResponse(w, r.URL.String(), "error decoding request body")
+			return
+		}
+		param = req.IndicesOrPubKey
+	}
+
+	queryIndices, err := parseValidatorParamsToIndices(param, 100)
```
pk910, line 64:
```suggestion
	pubkeys, err := parseValidatorParamsToPubkeys(param, 100)
```

handlers/api/validator_deposits_v1.go
```diff
@@ -0,0 +1,121 @@
...
+		Indices: queryIndices,
+	}, true)
+
+	pubkeys := make([][]byte, 0, len(relevantValidators))
+	for _, relevantValidator := range relevantValidators {
+		pubkey := relevantValidator.Validator.PublicKey[:]
+
+		pubkeys = append(pubkeys, pubkey)
+	}
+
```
pk910, line 80:
```suggestion

```

handlers/api/validator_deposits_v1.go
```diff
@@ -0,0 +1,121 @@
...
+// @Router /api/v1/validator/{indexOrPubkey}/deposits [get]
+func ApiValidatorDepositsV1(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+
+	vars := mux.Vars(r)
+	var param string
+	if r.Method == http.MethodGet {
+		// Get the validators from the URL
+		param = vars["indexOrPubkey"]
+	} else {
```
pk910, line 49: this codepath is never hit as the endpoint accepts GET requests only.
can probably remove the whole if else block and use `param := vars["indexOrPubkey"]`

pk910 (changes requested):

handlers/api/general.go
pk910: added a second helper function `parseValidatorParamsToPubkeys`, to parse the params to a list of pubkeys and only lookup the ones specified by index

remyroy:
Very good points. Let me take some time to work out the fix and check your suggested changes.

handlers/api/validator_deposits_v1.go
remyroy: I've removed the else part.

pk910 (approved):

pk910:
Looks good :) 
Thank you very much 👍

PR #340, opened 2025-04-29
fix: set ClientsCLPageData node ENR from node identity

### LINKED ISSUE #339: fulu-support: Consensus clients page displays empty ENRs
I'm running local peerDAS devnets and noticing that the 'Consensus clients' page shows empty ENRs for some of my nodes even when I'm able to see the ENR when querying 'node/identity' directly.

Image

hughy:
use the node's identity from 'node/identity' to set ENR if the ENR earlier returned from a peer (i.e., 'node/peers') is empty or outdated

fix empty ENR on Consensus clients page for nodes with ENR available at 'node/identity'

Closes #339 

image

### DIFF
--- a/handlers/clients_cl.go
+++ b/handlers/clients_cl.go
@@ -258,21 +258,20 @@ func buildCLClientsPageData() (*models.ClientsCLPageData, time.Duration) {
 				Alias:  client.GetName(),
 				Type:   "internal",
 			}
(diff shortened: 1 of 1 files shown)


pk910 (approved):

pk910:
Heya @hughy,
thanks for the fix :) looks good

Issue #327, opened 2025-04-17
Add Validator Top Ups


coincashew:
Add new option:
Validator > Submit Top Ups

wen? please

Thanks!

ohko4711:
request too!
also I think we can add upgrade  type 0x1 to type 0x2, I think is just sepcial case for consolidation requests(source pubkey==tartget pubkey), This way, users don't need to perform this operation on an additional platform(eg. staking launchpad)

ohko4711:
> request too!  也请求了！ also I think we can add upgrade type 0x1 to type 0x2, I think is just sepcial case for consolidation requests(source pubkey==tartget pubkey), This way, users don't need to perform this operation on an additional platform(eg. staking launchpad)此外，我认为我们可以将升级类型 0x1 添加到类型 0x2，我认为这只是合并请求的特殊情况（源公钥 == 目标公钥）。这样，用户无需在其他平台（例如，staking launchpad）上执行此操作。

I'm sorry, I see it's already supported, but I didn't see it in the target dropdown menu, which led me to mistakenly believe it wasn't supported.


![Image](https://github.com/user-attachments/assets/2b10dc50-19ee-4ae1-95d0-d5d0da2db88d)

pk910:
both fixed by #358

PR #358, opened 2025-05-20
add validator topup form & enhance ux of consolidation target selection

pk910:
This PR adds a validator topup form that allows topping up arbitrary validators:
image

In addition, this PR improves the validator selector for arbitrary target validators:
image
Instead of fully relying on the search input, the control now shows the validators owned by the current wallet by default.

### DIFF
--- a/handlers/submit_deposit.go
+++ b/handlers/submit_deposit.go
@@ -1,12 +1,17 @@
 package handlers
 
 import (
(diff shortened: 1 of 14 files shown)


pk910:
fixes #327

barnabasbusa (approved):

PR #376, opened 2025-06-10
feat: add invert button for filter

barnabasbusa:
Screenshot 2025-06-10 at 15 41 45

### DIFF
--- a/handlers/slots_filtered.go
+++ b/handlers/slots_filtered.go
@@ -136,12 +136,6 @@ func buildFilteredSlotsPageData(pageIdx uint64, pageSize uint64, graffiti string
 	if pname != "" {
 		filterArgs.Add("f.pname", pname)
 	}
-	if invertgraffiti {
-		filterArgs.Add("f.ginvert", "on")
-	}
-	if invertextradata {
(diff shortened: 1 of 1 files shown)


mattevans (approved):

handlers/slots_filtered.go
```diff
@@ -103,21 +115,36 @@ func getFilteredSlotsPageData(pageIdx uint64, pageSize uint64, graffiti string,
...
 	}
 	if pname != "" {
 		filterArgs.Add("f.pname", pname)
 	}
+	if invertgraffiti {
+		filterArgs.Add("f.ginvert", "on")
+	}
+	if invertextradata {
+		filterArgs.Add("f.einvert", "on")
+	}
```
pk910, line 144: looks like `invertgraffiti` & `invertextradata` got duplicated here, so you'll find the filter twice in the generated urls ;)

pk910 (changes requested):

handlers/slots_filtered.go
barnabasbusa: Fixed.

pk910 (approved):

PR #402, opened 2025-06-28
refactor snooper client & ignore proposer execution times

pk910:


pk910:
added colored rendering for execution time:
image

>= 500: yellow
>= 1000: red

mattevans (approved):

PR #410, opened 2025-06-30
fix jump paging controls

pk910:
retain filters when jumping to a specific page


mattevans (approved):

pk910:
> Just checking the hidden `f` inputs are needed? Wasn't sure where they tie in.
> 
> LGTM though!

yea, the page handler only processes filters when "f" is included in the url.
That's been a easy approach to skip the whole filter logic in an earlier version with no filters applied :D  but is actually not used anymore,   but the page handler still checks for it

PR #467, opened 2025-09-05
Add el chain config handling, eth_config checks + visualization & blob base fee calculation

pk910:
supersedes #440

eth_Config in el clients view:
image

blob fee properties in slot details:
image


pk910:
I've added a visualization for spec/config mismatches:
image

image

PR #634, opened 2026-04-14
Reduce DB query build overhead with shared helpers

Sahil-4555:
This PR reduces DB query build overhead by centralizing placeholder construction and fork‑ID filtering into shared helpers, then applying them across hot query paths. The changes remove repeated string building and temporary slice allocations in many `IN (...)` filters and batch queries, while also speeding up byte‑slice map key generation for lookup maps. Benchmarks show the placeholder builder is about 4.4x faster (18,568 ns/op down to 4,169 ns/op) with much fewer allocations (131 to 37), and byte‑slice key conversion is about 1.7x faster (209 ns/op down to 121 ns/op). Overall, the query assembly paths are now leaner and faster, especially when filters include large ID lists.

### DIFF
--- a/db/common.go
+++ b/db/common.go
@@ -39,12 +39,9 @@ var logger = logrus.StandardLogger().WithField("module", "db")
 func checkDbConn(dbConn *sqlx.DB, dataBaseName string) {
 	// The golang sql driver does not properly implement PingContext
 	// therefore we use a timer to catch db connection timeouts
-	dbConnectionTimeout := time.NewTimer(15 * time.Second)
-
-	go func() {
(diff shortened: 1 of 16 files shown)


pk910:
Heya @Sahil-4555 
Looks all good :) I'll gonna accept it as it simplifies the code structure for db queries too.
But in general I'd like to avoid heavy optimization only changes in dora.
I'm pretty sure there is a lot to optimize,  but given the complexity of some code parts, I'd like to keep it readable and structured as is for maintainablility.

pk910 (approved):

PR #646, opened 2026-04-18
fix Redis bool cache key prefix

Sahil-4555:
## Summary

Fixes `RedisCache.GetBool` to use the configured Redis key prefix when reading values.

`SetBool` writes bool values using the prefixed key, but `GetBool` previously looked up the raw unprefixed key. This meant bool values stored through `SetBool` would not be found when a Redis prefix was configured.

## Changes

- Updated `GetBool` to read from the same prefixed key format used by `SetBool`.

### DIFF
--- a/cache/redis_cache.go
+++ b/cache/redis_cache.go
@@ -72,7 +72,7 @@ func (cache *RedisCache) SetBool(ctx context.Context, key string, value bool, ex
 
 func (cache *RedisCache) GetBool(ctx context.Context, key string) (bool, error) {
 
-	value, err := cache.redisRemoteCache.Get(ctx, key).Result()
+	value, err := cache.redisRemoteCache.Get(ctx, fmt.Sprintf("%s%s", cache.keyPrefix, key)).Result()
 	if err != nil {
 		return false, err
 	}


cache/redis_cache.go
```diff
@@ -67,12 +66,12 @@ func (cache *RedisCache) GetUint64(ctx context.Context, key string) (uint64, err
...
 
 func (cache *RedisCache) SetBool(ctx context.Context, key string, value bool, expiration time.Duration) error {
-	return cache.redisRemoteCache.Set(ctx, fmt.Sprintf("%s%s", cache.keyPrefix, key), fmt.Sprintf("%t", value), expiration).Err()
+	return cache.redisRemoteCache.Set(ctx, cache.keyPrefix+key, strconv.FormatBool(value), expiration).Err()
 }
 
 func (cache *RedisCache) GetBool(ctx context.Context, key string) (bool, error) {
 
-	value, err := cache.redisRemoteCache.Get(ctx, key).Result()
+	value, err := cache.redisRemoteCache.Get(ctx, cache.keyPrefix+key).Result()
```
pk910: nice find. this was actually a bug as the old `GetBool` key did not match the `SetBool` key.

pk910:
Heya @Sahil-4555

Thanks, I went through all the changes. They're technically clean.

Most of this is micro-optimization though (sprintf -> string concat, Replace -> TrimPrefix, %x -> hex.EncodeToString, bytes.Equal, ...) and I don't think it's worth merging just for that. The wins are in the ns range, but these all run inside page handlers that take milliseconds for processing and db queries anyway, so it won't make any measurable difference.

Two things did stand out as actual fixes rather than perf:

- GetBool was missing the key prefix, so it would never match what SetBool writes. There's no caller for it in the app right now so it has no real impact today, but it's still a valid fix.
- The re-check in isSszCompatible after taking the write lock is fine too. Just a note: it's not really a race, the map is already behind the mutex - it just avoids running ValidateType twice when two goroutines hit the write lock at the same time.

Could you pull the GetBool fix out into its own small PR? I'd happily merge that. The rest I'd rather leave out to keep the churn down.

Sahil-4555:
> Heya @Sahil-4555
> 
> Thanks, I went through all the changes. They're technically clean.
> 
> Most of this is micro-optimization though (sprintf -> string concat, Replace -> TrimPrefix, %x -> hex.EncodeToString, bytes.Equal, ...) and I don't think it's worth merging just for that. The wins are in the ns range, but these all run inside page handlers that take milliseconds for processing and db queries anyway, so it won't make any measurable difference.
> 
> Two things did stand out as actual fixes rather than perf:
> 
> * GetBool was missing the key prefix, so it would never match what SetBool writes. There's no caller for it in the app right now so it has no real impact today, but it's still a valid fix.
> * The re-check in isSszCompatible after taking the write lock is fine too. Just a note: it's not really a race, the map is already behind the mutex - it just avoids running ValidateType twice when two goroutines hit the write lock at the same time.
> 
> Could you pull the GetBool fix out into its own small PR? I'd happily merge that. The rest I'd rather leave out to keep the churn down.

Done. let me know if there are any other feedback points.

pk910 (approved):

PR #651, opened 2026-04-21
fix: only apply beaconapi.clientIndex filter when explicitly set

barnabasbusa:
## Summary
- `beaconapi.ClientIndex` was typed `int`, so its Go zero value (`0`) always satisfied the `>= 0` check in `utils.ReadConfig`, collapsing `beaconapi.endpoints` down to `endpoints[0]` for users who never set `clientIndex`. Multi-endpoint setups only ever exposed the first beacon client on `/clients/consensus` and in the indexer.
- Change `ClientIndex` to `*int` so `nil` means "unset"; the single-endpoint selection now only runs when the field is explicitly configured (YAML or `BEACONAPI_CLIENT_INDEX`).
- Introduced in 11f18e66 (Nov 2025).

## Test plan
- [x] Verified locally: config with two `beaconapi.endpoints` and no `clientIndex` previously rendered only the first client on `/clients/consensus`; after the fix, both clients are listed and indexed.
- [x] `go build ./...` passes.
- [ ] Still honors explicit `clientIndex: N` (selects a single endpoint) and `BEACONAPI_CLIENT_INDEX` env var.

### DIFF
--- a/types/config.go
+++ b/types/config.go
@@ -97,7 +97,7 @@ type Config struct {
 		Endpoint     string           `yaml:"endpoint" envconfig:"BEACONAPI_ENDPOINT"`
 		Endpoints    []EndpointConfig `yaml:"endpoints"`
 		EndpointsURL string           `yaml:"endpointsUrl" envconfig:"BEACONAPI_ENDPOINTS_URL"`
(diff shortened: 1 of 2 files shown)


barnabasbusa:
reopening from origin with correct branch prefix (bbusa/)

PR #672, opened 2026-05-01
fix: keep Dora indexing across the Heze fork boundary

lodekeeper:
## Summary

Fix Dora so it keeps indexing through the Heze fork boundary instead of breaking at slot 32.

This PR does two things:

1. Forces Dora's consensus beacon client onto JSON requests instead of SSZ for now.
2. Adds a narrow Heze fallback in `GetBlockBodyByBlockroot()` for the exact signed-block-contents SSZ decode failure, plus regression coverage.

## Why

With the updated published images and a simple Lodestar+Besu Kurtosis devnet, Dora indexed blocks normally through slot 31, then failed exactly at the Heze fork boundary with:

- `failed to decode heze signed block contents`
- `Message.Body.ProposerSlashings:o: incorrect offset: first offset 396 does not match expected 392`

Source-level diagnosis:

- Dora pins `github.com/ethpandaops/go-eth2-client v0.1.1`
- that path still expects SSZ `SignedBeaconBlock` / related SSZ types where the Heze-era APIs now expose block contents / state / envelope shapes that it cannot decode correctly
- forcing JSON requests avoids those broken SSZ decode paths

### Repro

Using updated published images with `--image-download always`:

- `ethpandaops/besu:focil`
- `ethpandaops/lodestar:focil`
- Heze at epoch 1

### Before

- Dora broke at slot 32
- `/blocks` stopped at the Heze boundary
- logs showed the Heze signed-block-contents decode failure above

### After

With the patched image (`nflaig/dora:heze-slot32-fix`):

- Dora backfilled well past Heze on the same live chain
- `/blocks` showed slots deep into Heze (`slot/224`, `223`, ...)
- the Heze/Gloas decode errors no longer appeared in Dora logs under the stock Kurtosis config

### Tests

- Added a regression test for the exact Heze matcher error text
- Added a real slot-32 Heze JSON fixture test for the JSON block-body helper

## Notes

This is an intentionally pragmatic workaround until `go-eth2-client` gains full Heze/Gloas SSZ contents/state support.


pk910:
superseded by #673

PR #684, opened 2026-05-07
feat(slot): regroup slot sub-tabs and rename slugs

barnabasbusa:
## Summary
The slot details bar previously exposed every consensus / execution operation as its own tab, which got long once a block carried both legacy and EL-request variants of the same operation. This PR collapses near-duplicates, shortens the slugs, and reorders the bar to follow what the user typically cares about.

URL hash behavior is unchanged from master — direct visits to \`/slot/<id>#<top-level-tab>\` still activate the tab via the existing simple \`location.hash\` guard. A follow-up PR will add full URL hash sync (\`replaceState\` on tab change, \`hashchange\` listener, sub-pill awareness).

## Tab regroup
**Combined tabs (single tab, stacked sections):**
- **Deposits** — Deposit Requests + Legacy Deposits
- **Withdrawals** — Withdrawal Requests + Executed Withdrawals
- **Slashings** — Attester + Proposer Slashings

**Transactions** is now one top-level tab with an inner Bootstrap **pills** sub-nav: \`Transactions\` (default) and \`Access List\`.

**Slug renames (URL hash = pane id, single source of truth in the template):**
| Old | New |
|---|---|
| \`ptcVotes\` | \`ptc\` |
| \`blsChange\` | \`bls\` |
| \`blobSidecars\` | \`blobs\` |
| \`inclusionLists\` | \`focils\` |
| \`consolidationRequests\` | \`consolidations\` |
| \`voluntary-exits\` | \`exits\` |
| \`attester-slashings\` + \`proposer-slashings\` | \`slashings\` |
| \`depositRequests\` (folded into) | \`deposits\` |
| \`withdrawalRequests\` (folded into) | \`withdrawals\` |
| \`accessList\` (sub-pill of) | \`bals\` |

**New tab order:** Overview, Transactions (with inner Tx / Access List pills), Bids, Attestations, Deposits, Consolidations, Exits, Withdrawals, Slashings, PTC, BLS, Blobs, FOCILs, Proofs, Download.

The blob loader link (\`templates/slot/blobs.html\`) now points at \`#blobs\`.

## bytesToHex collision fix
\`templates/slot/attestations.html\` defines a different (simpler) global \`bytesToHex\` that doesn't decode base64. After regrouping, the BAL pane is now emitted before the attestations pane, so the simpler version overrides the base64-aware one and the BAL transaction popup ended up rendering Hash/From/To with a \`0\` prefix per base64 char. Renamed the BAL helper to \`balBytesToHex\` so the two coexist regardless of emission order.

## Test plan
- [ ] On a block with both transactions and a BAL, the **Transactions** tab shows the inner pills; clicking each pill switches sections.
- [ ] On a block with both attester and proposer slashings, the **Slashings** tab shows both sections stacked with their own headers.
- [ ] On a Pectra block with both EL deposit requests and legacy deposits, the **Deposits** tab shows both sections; same for **Withdrawals**.
- [ ] BAL transaction popup shows valid hex Hash/From/To (no more \`0R0B...\` garbage).
- [ ] Visiting \`/slot/<id>#bids\`, \`/slot/<id>#blobs\`, \`/slot/<id>#download\` directly still activates the matching top-level tab.
- [ ] Other top-level tabs (Attestations, PTC, BLS, FOCILs, Proofs) render with the new slugs.

### DIFF
--- a/templates/slot/blobs.html
+++ b/templates/slot/blobs.html
@@ -32,7 +32,7 @@
           <div class="blobloader-container" data-blob-idx="{{ $blob.Index }}">
             <div class="row border-bottom p-1 mx-0">
               <div class="col text-center">
-                <a class="btn btn-primary blobloader-button" href="?blob={{ $blob.Index }}#blobSidecars" role="button">Load Blob Data</a>
+                <a class="btn btn-primary blobloader-button" href="?blob={{ $blob.Index }}#blobs" role="button">Load Blob Data</a>
               </div>
             </div>
           </div>
--- a/templates/slot/block_access_list.html
+++ b/templates/slot/block_access_list.html
@@ -1828,18 +1828,18 @@
           </div>
           <div class="tx-callout-row">
             <span class="tx-callout-label">Hash:</span>
-            <span class="tx-callout-value">0x${bytesToHex(tx.hash)}</span>
-            <i class="fa fa-copy tx-callout-copy" onclick="copyToClipboard('0x${bytesToHex(tx.hash)}')" title="Copy to clipboard"></i>
+            <span class="tx-callout-value">0x${balBytesToHex(tx.hash)}</span>
+            <i class="fa fa-copy tx-callout-copy" onclick="copyToClipboard('0x${balBytesToHex(tx.hash)}')" title="Copy to clipboard"></i>
           </div>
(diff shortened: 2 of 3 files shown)


barnabasbusa:
probably a breaking change thats not really needed

PR #701, opened 2026-05-19
fix: attach exec times to blocks after gloas payload arrives

barnabasbusa:
## Summary

- Exec Time column was empty for every post-gloas slot because the snooper could never match its `engine_newPayloadV5` timings to the corresponding beacon block.
- Root cause: in EIP-7732 the beacon block body no longer carries the execution payload. The body arrives first on its own gossip topic, so `processBlock` runs `addBlockToExecBlockMap` while `blockIndex.ExecutionHash` is still zero, and the zero-hash guard skips it. When the payload later lands via `execution_payload_available`, `setBlockIndex` populates the hash but `addBlockToExecBlockMap` is never called again — the block stays absent from `execBlockMap`, and the snooper's `GetBlocksByExecutionBlockHash` lookup always misses. The exec time gets parked in the 5-minute fallback cache and expires unused.
- Fix: in `processExecutionPayloadAvailableEvent`, once `EnsureExecutionPayload` has populated `ExecutionHash`, call `addBlockToExecBlockMap(block)` so the block becomes findable, and drain any exec times the snooper cached for that hash while the block was missing.
- The drain-from-cache logic is factored out of `processBlock` into `drainCachedExecutionTimes` so both code paths share it.

## Test plan

- [x] Build & vet clean (`go build ./... && go vet ./...`)
- [x] Run a kurtosis devnet with `gloas_fork_epoch: 1` and `snooper_enabled: true`, confirm the Exec Time column populates for slots in epoch 1+ (previously only epoch 0 had values)
- [x] Confirm pre-gloas behaviour unchanged (epoch 0 still shows exec times)
- [x] Verify no regressions in the per-block detail view's exec-time breakdown

### DIFF
--- a/indexer/beacon/client.go
+++ b/indexer/beacon/client.go
@@ -460,31 +460,7 @@ func (c *Client) processBlock(slot phase0.Slot, root phase0.Root, header *phase0
 		c.indexer.blockCache.addBlockToParentMap(block)
 		c.indexer.blockCache.addBlockToExecBlockMap(block)
 
-		// Check for cached execution times and add them to the block
-		blockIndex := block.GetBlockIndex(c.indexer.ctx)
-		if blockIndex != nil && !bytes.Equal(blockIndex.ExecutionHash[:], zeroHash[:]) {
-			executionHash := common.Hash(blockIndex.ExecutionHash)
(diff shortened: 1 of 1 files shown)


barnabasbusa:
Closing in favor of #700

PR #728, opened 2026-06-05
resolve builder names + external links from buildoor inventory

barnabasbusa:
Screenshot 2026-06-05 at 16 29 05

## Summary
- Adds a `BuildoorInventory` service that polls one or more [buildoor](https://github.com/ethpandaops/buildoor) instances, maps builder pubkeys → `{name, url}` via the indexer pubkey cache, and exposes the result to the existing `GetValidatorName(idx | BuilderIndexFlag)` lookup path plus a new `GetBuilderURL(idx)`.
- Two configurations under `frontend:` — `buildoorOverviewUrl` (a multi-instance overview server; dora enumerates via `/api/overview/hosts`) or `buildoorUrls` (an explicit list of per-instance URLs). Both paths converge on `<instance>/api/buildoor/overview`. `buildoorUrls` wins if both are set.
- Plumbs `BuilderURL` through `SlotPageBid`, `SlotPagePayloadHeader`, `SlotsPageDataSlot`, `SlotsFilteredPageDataSlot`, `BlocksPageDataBlock` and the 5 templates that render builders. New `FormatBuilderWithURL` / `FormatBuilderWithIndexAndURL` helpers append a small `fa-external-link-alt` icon next to the name when a URL is known — existing helper signatures are unchanged so validator-context templates that reuse `formatBuilder` aren't touched.

## Caching
The per-instance HTTP fan-out runs only inside `BuildoorInventory.refresh()`, invoked by the background `runUpdaterLoop` on `buildoorRefreshInterval` (default 5m). Page handlers go through `GetValidatorName` / `GetBuilderURL`, which is a mutex-locked map lookup — zero HTTP per page render.

## Example config

```yaml
frontend:
  # one of:
  buildoorOverviewUrl: "https://buildoor.glamsterdam-devnet-5.ethpandaops.io"
  # or:
  buildoorUrls:
    - "https://api-buildoor-prysm-ethrex-1.srv.glamsterdam-devnet-5.ethpandaops.io"
  buildoorRefreshInterval: 5m
```

Devnet wiring lands separately:
- ethereum-package (kurtosis): https://github.com/ethpandaops/ethereum-package/pull/1413
- ansible-collection-general (helm): https://github.com/ethpandaops/ansible-collection-general/pull/542

Visual: the Builder column still links to dora's `/builder/<idx>`. When the inventory resolves a builder, a small external-link icon appears immediately after the name, opening the buildoor instance in a new tab.

## Test plan
- [ ] On glamsterdam-devnet-5: set `buildoorOverviewUrl: "https://buildoor.glamsterdam-devnet-5.ethpandaops.io"`, open a slot with bids, and confirm the Builder column shows the parsed name (e.g. `prysm-ethrex-1`) with an external-link icon that opens `https://api-buildoor-prysm-ethrex-1.srv.glamsterdam-devnet-5.ethpandaops.io`.
- [ ] Same devnet, switch to `buildoorUrls: [<instance1>, <instance2>]` and confirm identical behavior.
- [ ] With no buildoor config set, builder names + URLs are empty; existing templates render unchanged.
- [ ] Watch dora logs for `buildoor inventory refreshed: N resolved, M unresolved` on the configured refresh interval.


services/buildoor.go
```diff
@@ -0,0 +1,272 @@
...
+			continue
+		}
+
+		flaggedIdx, found := b.beaconIndexer.GetValidatorIndexByPubkey(pubkey)
+		if !found {
+			unresolvedCount++
+			continue
+		}
+
+		rawIdx := uint64(flaggedIdx) &^ beacon.BuilderIndexFlag
```
pk910: the builder index is part of the overview response too.
overview.builder_index (gated on overview.is_registered) should be more accurate than relying on the validator set cache (which is delayed by the next epoch boundary)

barnabasbusa:
Good call — fixed in a1c8f5a. Now gates on `is_registered` and uses `overview.builder_index` directly; drops the pubkey decode + indexer cache lookup entirely from the resolve path.

barnabasbusa:
Re: the `GetBuilderURL` flag-mismatch flags — `GetBuilderURL` and `GetValidatorName` look up different namespaces, so they intentionally took different keys:

- `GetValidatorName(idx | BuilderIndexFlag)` is polymorphic across the validator + buildoor namespaces; the flag bit tells ChainService which to consult.
- `GetBuilderURL(rawIdx)` only ever looks at the buildoor namespace, so no flag was needed.

All five `GetBuilderURL` call sites pass the raw index, matching how the inventory stores entries — no actual data loss.

That said, the asymmetric calling convention is a real footgun. Fixed in ec47314 — the lookups now strip the flag internally, so either form works.

Re: the nil-guard on `GetBuilderURL` — it's present at services/buildoor.go:235 (`if b == nil { return "" }`), mirror of the `GetBuilderName` guard you cited.

pk910 (approved):

PR #749, opened 2026-06-21
fix: prevent double panic in recovery handlers when panic value is not an error type

Sahil-4555:
Found a small but important issue in our panic recovery handlers across the codebase. Currently all recovery handlers are doing `err.(error)` type assertion on the recovered value. The problem is that `recover()` returns `interface{}` and in Go, panic can be called with any type - string, int, etc. If some third party library does `panic("some message")` instead of `panic(errors.New("some message"))`, then our `err.(error)` assertion will panic again inside the recovery handler itself. This means the deferred restart (`go func.restart()`) will never execute and that goroutine will be permanently dead. Basically the recovery handler defeats its own purpose.

The fix is straightforward - replaced `err.(error)` with `fmt.Errorf("%v", err)` which safely converts any panic value to an error. No behaviour change at all, same log messages, same restart logic, just the error wrapping is now safe for all types.

Also consolidated all the inline `defer func() { if err := recover()...` blocks across 9 files to use the shared `utils.HandleSubroutinePanic` helper instead. This removes a lot of duplicated code and keeps things consistent. The helper already handles everything — safe error wrapping, logging with identifier, and optional restart after 10 seconds delay.


pk910:
Heya @Sahil-4555 
thx for the fix,  looks good to me.
One small suggestions as you're already touching the panic handlers:
could you please replace all the 
```
defer func() {
	if err := recover(); err != nil {
	...	
```
blocks by a simple 
`defer utils.HandleSubroutinePanic("identifier", subroutineFn)`

so the helper is used consistently :)

Sahil-4555:
> Heya @Sahil-4555 thx for the fix, looks good to me. One small suggestions as you're already touching the panic handlers: could you please replace all the
> 
> ```
> defer func() {
> 	if err := recover(); err != nil {
> 	...	
> ```
> 
> blocks by a simple `defer utils.HandleSubroutinePanic("identifier", subroutineFn)`
> 
> so the helper is used consistently :)

Done.

pk910 (approved):

PR #758, opened 2026-06-23
feat(dora): show EIP-7778 block gas delta in slot overview

qu0b:
## Summary

Adds a **Block Gas Delta** row below *Gas Used* in the slot execution payload section, showing the difference between `block.gasUsed` and `Σ receipt.gasUsed`.

Amsterdam (EIP-7778) changed `block.gasUsed` from the naive sum of receipts to `max(Σ regular, Σ state)`, introducing a 2D gas model where the block header and the receipts can diverge:

- **Positive delta** (`block.gasUsed > Σ receipts`): regular gas dominates. The gap equals EIP-3529 gas refunds when no state gas is present, or refunds minus state gas in mixed blocks. Senders were charged more in receipts than the block header reports.
- **Zero delta**: no refunds and no state gas — block and receipts agree (e.g. pure ETH transfer blocks).
- **Negative delta** (`block.gasUsed < Σ receipts`): state gas (EIP-8037) dominates. Receipts include both regular + state gas per tx, but the block header only records `max(regular, state)`, so state-heavy blocks have receipts that exceed `block.gasUsed`.

The row is only shown when the EL indexer has enriched **all** tx receipts in the block (guard: `elDataCount == len(transactions)`). The model struct exposes `BlockGasDelta` (absolute value), `BlockGasDeltaExcess` (sign), and `SumTxGasUsed` for the receipts total.

## Screenshot

Block Gas Delta on glamsterdam-devnet-6 Kurtosis network

*Pure ETH-transfer block: delta = +0, confirming block header and receipts agree when there are no refunds and no state gas writes.*

## Test plan

- [x] Deploy against a glamsterdam-devnet-6 network — **Block Gas Delta** row appears below Gas Used for blocks with EL-indexed transactions
- [x] Zero-delta block (simple transfers): shows `+0 [EIP-7778 ▲]` with receipts sum matching `block.gasUsed`
- [ ] Block with SSTORE refunds: delta shows a positive non-zero value equal to the sum of EIP-3529 refunds
- [ ] Block with heavy state gas writes (EIP-8037): delta goes negative, `EIP-7778 ▼` badge appears
- [ ] Block where EL indexer has not yet enriched all receipts: row does not appear

### DIFF
--- a/handlers/slot.go
+++ b/handlers/slot.go
@@ -866,8 +866,10 @@ func getSlotPageBlockData(ctx context.Context, blockData *services.CombinedBlock
 			getSlotPageTransactions(ctx, pageData, executionPayload.Transactions, blockUid)
 		}
 
-		// EIP-7778: compute gas refund delta = block.gasUsed - sum(receipt.gasUsed).
-		// block.gasUsed is pre-refund (EIP-7778); each receipt.gasUsed is post-refund.
+		// EIP-7778: compute gas refund delta = |block.gasUsed - sum(receipt.gasUsed)|.
+		// In Amsterdam these two counters ALWAYS diverge (they were equal pre-Amsterdam):
+		//   block.gasUsed  = gp.Used()           = max(sum_regular, sum_state)
+		//   sum(receipts)  = gp.CumulativeUsed() = sum(max(pre_refund-refund, floor))
 		// Only computed when EL indexing has enriched ALL tx receipts in this block.
 		if len(pageData.Transactions) > 0 {
 			var sumTxGas uint64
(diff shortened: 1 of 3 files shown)


qu0b:
## glamsterdam-devnet-6 test: EIP-7843 slot_number + EIP-7928 BAL hash + EIP-7778 gas delta ✅

Tested on glamsterdam-devnet-6. All three execution payload fields are visible in the slot overview:

EIP-7843 Slot Number, EIP-7778 Gas Refund Delta, EIP-7928 BAL Hash

**EIP-7843: Slot Number** — The `Slot Number: 2864` row in the Execution Payload section shows the beacon slot number with its EIP badge.

**EIP-7778: Gas Refund Delta** — The `Gas Refund Delta: +0 EIP-7778 ▲ (receipts sum: 300912)` row shows the difference between block-level `gasUsed` and the sum of individual receipt `gasUsed` values, with an arrow indicating direction.

**EIP-7928: BAL Hash** — The `BAL Hash` row shows the keccak256 of the RLP-encoded Block Access List with its EIP badge.

pk910 (approved):

PR #762, opened 2026-06-25
fix: safe type assertion in beacon config and align rpc proxy IP extraction with middleware

Sahil-4555:
Hey,

Two small but important fixes in this PR.

**1. Safe type assertion in GetConfigSpecs (beaconapi.go)**
In `GetConfigSpecs`, we were doing an unchecked type assertion on the beacon node config response - `specs["data"].(map[string]interface{})`. The nil check only checked if `data` is missing, but if the beacon node returns `data` as a string or array, it would panic and crash the explorer. Since this is called during client initialization, a panic here crashes the whole explorer.
Fixed it by using a proper comma-ok type assertion. If it is not a map, we now return a clean error instead of panicking. No change in behavior for normal responses.

**2. Centralized proxyCount configuration and consistent IP extraction (rpcproxy.go & types/config.go)**
The RPC proxy had its own `getClientIP` method which was not aligned with the middleware's `GetClientIP`. The RPC proxy was taking the leftmost IP from `X-Forwarded-For` which is easily spoofable by the client.
* Aligned the RPC proxy to use the shared `middleware.GetClientIP` instead.
* As suggested, we moved the `proxyCount` setting to the `server` config section (`server.proxyCount`) since it is a base connectivity option that affects multiple components.
* Added fallback logic: if `server.proxyCount` is not configured, it will fall back to using the deprecated `rateLimit.proxyCount` so we don't break old setups. Commented the old field as deprecated.
* Initialized the call rate limiter in `main.go` using the same resolved proxy count.

### DIFF
--- a/clients/consensus/rpc/beaconapi.go
+++ b/clients/consensus/rpc/beaconapi.go
@@ -295,11 +295,12 @@ func (bc *BeaconClient) GetConfigSpecs(ctx context.Context) (map[string]interfac
 		return nil, fmt.Errorf("error retrieving specs: %v", err)
 	}
 
-	if specs["data"] == nil {
-		return nil, fmt.Errorf("specs data is nil")
+	dataMap, ok := specs["data"].(map[string]interface{})
+	if !ok {
+		return nil, fmt.Errorf("specs data is not an object (got %T)", specs["data"])
 	}
 
-	parsedSpecs := utils.ParseSpecMap(specs["data"].(map[string]interface{}))
+	parsedSpecs := utils.ParseSpecMap(dataMap)
 
 	return parsedSpecs, nil
 }
--- a/cmd/dora-explorer/main.go
+++ b/cmd/dora-explorer/main.go
@@ -90,7 +90,7 @@ func main() {
 	}
 
 	if cfg.RateLimit.Enabled {
-		err = services.StartCallRateLimiter(cfg.RateLimit.ProxyCount, cfg.RateLimit.Rate, cfg.RateLimit.Burst)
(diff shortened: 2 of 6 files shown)


pk910:
Heya @Sahil-4555
I can confirm both issues, and the fix for the first looks good.
I think the second fix needs more work. Re-using the proper `middleware.GetClientIP` sounds good,  but it uses the `ProxyCount` setting from the ratelimit config section (`rateLimit.proxyCount`). 
That whole section is usually undefined if the rate limiter is not configured, so relying on a setting from there for another code part feels wrong.

I'd instead move the `ProxyCount` setting somewhere more central, as it affects more than only the rate limiter.
I think the server config section would fit better as it kinda belongs to the base connectivity options..

The cherry on top would be if you can somehow still use the old setting at `rateLimit.proxyCount` if the new `server.proxyCount` is undefined - so we don'T have a breaking config change (but comment the old setting as deprecated) :)

Sahil-4555:
hey, @pk910 

I have updated the PR as per your suggestion:
* Moved the `proxyCount` to `server.proxyCount` since it's a general connection option.
* Added the fallback option: if `server.proxyCount` is not defined, it will pick the value from the old `rateLimit.proxyCount`. Also added a comment marking the old one as deprecated.
* Both the rate limiter and the client IP middleware now use this resolved value.
* I have also written a quick unit test in `types/config_test.go` to test all fallback cases, and everything is passing fine.

Please check the https://github.com/ethpandaops/dora/pull/762/changes/3cc9b43ab893104d601fc2225343f26a05eefad8  commits.

pk910 (approved):

pk910:
pls run `go fmt ./...` and commit fixes

PR #768, opened 2026-06-26
EL explorer: compaction + filterable, keyset-paginated views

pk910:
# EL explorer: standard pages, keyset pagination, and a smaller/longer-lived store

## Summary

This branch rounds out the execution-layer (EL) explorer and reworks how its data
is stored and retained. It has four threads that build on each other:

1. **More explorer surface** — the standard EIP-3091 pages (tokens, token detail,
   transactions, transfers), keyset (boundary) pagination everywhere, collapsible
   filter panels, a contract-bytecode tab, and `deploy` method detection.
2. **Less storage** — drops the per-event index table and redundant indexes, and
   moves by-hash lookups off a dedicated `tx_hash` index onto a compact prefix
   index (net reduction even after the new index).
3. **Tiered blockdb rework** — the `tiered` backend now actually stores exec data
   (previously a silent no-op), with configurable per-namespace cache eviction.
4. **Decoupled retention** — blockdb (cheap tier) can be retained far longer than
   the relational tables, and `/tx/{hash}` keeps working after the relational row
   is pruned by reconstructing the transaction from blockdb via a long-lived
   tx-hash index. Pruning state is persisted and surfaced in the UI.

### New EL explorer pages
- `/tokens`, `/token/<contract>` (token detail), `/transactions`, `/transfers`
  (global token-transfer list).
- Token detail page: keyset-paginated transfers + filter panel (slot, from, to,
  amount) + column selector.

### Address page
- **Keyset pagination on every tab** (transactions, internal txs, ERC20, NFT,
  block fees), replacing `OFFSET/LIMIT`. Withdrawals keep the shared offset pager
  (they come through the cache-merging beacon service).
- **Collapsible filter panels** on Transactions / ERC20 / NFT with a Direction
  (All / Out / In) selector plus slot, value/amount, gas, tip, type and status
  filters, applied via AJAX with in-place URL update (`history.replaceState`) so
  a filtered view is copy-pasteable.
- **Contract tab** for contract addresses: deployed bytecode via `eth_getCode`.
- Token-page link for token contracts; merged inclusion+execution status badge.

### Tiered blockdb rework
- `TieredEngine` now implements the exec-data capability (write-through S3 +
  pebble cache, read-through with S3 fallback). Previously the tiered backend
  silently dropped exec data, disabling all EL detail tabs.
- **Per-namespace retention**: the four unused per-block-section retention configs
  are collapsed into a single `BlockRetention`, and `ExecDataRetention` /
  `DutiesRetention` are added. In tiered mode they evict from the pebble cache
  (authoritative copy stays in S3); **in pebble mode they delete authoritatively**,
  so block/duties data is actually cleared (previously nothing cleared it). Exec
  data in pebble mode is owned by `executionIndexer.detailsRetention`, so a pebble
  `execDataRetention` is ignored with a warning. A `time→slot` resolver is
  injected for age eviction of slot-keyed namespaces.

### Decoupled retention + tx-hash index
- `ExecutionIndexer.DetailsRetention` lets blockdb exec data + the tx-hash index
  retain longer than the relational `Retention`; cleanup is split into a short
  relational cutoff and a long blockdb/index cutoff.
- **Long-lived tx-hash index** (10-byte prefix → `tx_uid`) drives by-hash lookups
  and `/tx/{hash}` after relational rows are pruned. Store is configurable via
  `ExecutionIndexer.TxHashIndexStore` (`auto` | `pebble` | `postgres`): native
  pebble namespaces (pinned, never cache-evicted) where a local pebble exists,
  or the relational `el_txhash` table (default for `s3`/`tiered`).
- `/tx/{hash}` reconstructs a pruned transaction from blockdb (envelope decoded
  from the payload, disambiguated by full hash; receipt fields from blockdb
  ReceiptMeta) — transparently, with no archive banner.
- Search-ahead partial-hash matching moves onto the prefix index (no seq scan
  after the `tx_hash` index is dropped).

### Pruning observability
- EL pruning status is persisted in `explorer_state` (relational/details
  pruned-before epochs + last 3 cleanup runs with per-object cleared counts and
  freed bytes) and exposed via the indexer.
- List views (`/transactions`, `/transfers`, address) show a small
  "epochs X–Y" availability badge when retention pruning is active (tail from the
  cleanup status, head read live since it advances independently).
- The tx not-found page notes when a hash may have aged out of the details
  retention window.
- Debug → Execution tab shows the pruning bounds + recent runs; Debug → Blockdb
  tab shows the tx-hash index store type and size (native size estimated over the
  ns4+ns5 namespaces that make up the index; relational via `el_txhash` size).

### Deployment detection
- Contract-creation txs render the method as **`deploy`** instead of mislabeling
  the init code's first 4 bytes. The signal is the index-time `tx.To() == nil`,
  stored as **bit 7 of `tx_type`** (`0x80`; EVM type in bits 0–6); the type
  filter expands each selected type to `{t, t|0x80}` so it stays a sargable `IN`.

### Shared components
- Keyset pager + offset pager driving the same UI; `el_filter_assets` helpers
  with a generalized bitmask multiselect (tx type, token type, column selector)
  and AJAX-submit + re-init hooks.

## Storage impact

The branch is a **net storage reduction** even with the new tx-hash index:

| Change | Direction | Rough magnitude |
|---|---|---|
| Drop `el_event_index` table (one row per emitted log + its `source`/`topic1` indexes) | **−** | the large win — GB-scale on a busy chain |
| Drop redundant `el_transactions_from_idx` / `to_idx` (composite `*_block_idx` cover them) | **−** | hundreds of MB |
| Drop the dedicated `el_transactions_tx_hash_idx`; switch PK to `tx_uid` (pgsql, via index promotion) | **−** | ~GB-scale on a busy chain |
| Add `event_count SMALLINT` badge column | **+** | ~2 bytes × row |
| Add the tx-hash prefix index (`el_txhash` table or native pebble ns4/ns5) | **+** | ~12 GB/90 d (Postgres) or ~5 GB/90 d (Pebble), operator-selectable |

Net: the relational set shrinks; the new index is a fraction of what it replaces
and is decoupled onto the cheap, longer-lived tier.

## Config additions
- `executionIndexer.detailsRetention` — longer retention for blockdb + tx-hash
  index (falls back to `retention`).
- `executionIndexer.txHashIndexStore` — `auto` | `pebble` | `postgres`.
- `blockDb.pebble.blockRetention` / `execDataRetention` / `dutiesRetention` —
  per-namespace retention (replaces the four unused per-section configs).
  Effective in both pebble mode (authoritative deletion) and tiered mode (cache
  eviction); `execDataRetention` applies in tiered mode only. The `tiered` engine
  reuses the top-level `blockDb.pebble` (cache) and `blockDb.s3` (primary)
  configs; there is no separate `tiered:` block.

## Migration / operational notes
- `20260625000000_el-index-rework` (one migration) **rewrites `el_transactions`**:
  backfills `event_count` and drops `el_event_index`; creates `el_txhash` and
  backfills the 10-byte prefix; drops the `tx_hash` and `from`/`to` indexes; and
  (Postgres only) promotes the unique `tx_uid` index to the primary key. SQLite
  keeps its `(block_uid, tx_hash)` PK (a swap there needs a full table rebuild,
  unnecessary for small deployments). Run in a maintenance window and `VACUUM`
  afterwards; the down-migration is unsupported.
- The **deploy** label and create-flag filtering only apply to rows indexed after
  the change; existing rows read back as non-creations until re-indexed.
- With `blockDb.engine: tiered`, exec data now requires this branch (it was a
  no-op before). `txHashIndexStore` defaults to `postgres` for `s3`/`tiered`
  (durable) and `pebble` for the `pebble` backend.

## Verification
- `go build ./...`, `go vet`, `gofmt` clean; `go test ./blockdb/...` passes
  (cache eviction, tx-hash index round-trip/range/prune, `HashPrefix`).
- All touched templates parse against the real funcmap.

## Refs
Replaces https://github.com/ethpandaops/dora/pull/739


pk910:
some screenshots of the new pages:

transactions list:
image

transfers:
image

tokens:
image

token details:
image

reworked address tabs with optional filters:
image

barnabasbusa (approved):

PR #777, opened 2026-06-29
fix: SQLite writer deadlock in retention cleanup and scrambled MEV proposal status

damilolaedwards:
## Summary

Two SQLite-specific bugs in the execution-data / MEV indexing paths. Both are silent on PostgreSQL but bite SQLite deployments (the default engine).

### 1. Retention cleanup deadlocks the writer on SQLite

`cleanupRetentionData` wrapped `DeleteElDataBeforeBlockUid` in a `RunDBTransaction`. That helper intentionally deletes in batches and commits between them, so it opens its own transactions. On SQLite `RunDBTransaction` takes the global writer mutex, so the inner per-batch transaction tried to re-lock it on the same goroutine , a plain `sync.Mutex`, so it deadlocks. Once that happens the mutex is never released and every other writer (finalization, synchronizer, validator/builder upserts, …) blocks behind it, so indexing stops until a restart.

This is reachable on a default config: the engine defaults to SQLite and retention defaults to 6 months, so any deployment with execution indexing enabled hits it once data ages past the cutoff on a cleanup tick.

Fix: let `DeleteElDataBeforeBlockUid` own its transactions and call it directly, and drop the unused `tx` parameter so it can't accidentally be nested again. PostgreSQL behaviour is unchanged (the outer transaction there only ever wrapped helpers that already committed independently).

### 2. MEV proposal status bound to the wrong values on SQLite

`UpdateMevBlockByEpoch` built the statement with the canonical-hash placeholders (`$3…`) appearing before the slot-range placeholders (`$1`/`$2`). SQLite's driver binds `$N` by the order each placeholder first appears in the text, not by the number, so the arguments were scrambled: the slot range compared against a hash, the hash list compared against an integer. The net effect is that canonical blocks were never marked `proposed`, and the per-epoch update silently touched the wrong rows.

Fix: number the placeholders in ascending order of appearance (hashes first, slot range last) with the argument slice in the same order. Works correctly on both engines.

## Tests

Added integration tests that run against a freshly migrated in-memory SQLite database:

- `TestUpdateMevBlockByEpoch` : verifies canonical, orphaned, and out-of-range blocks each get the correct `proposed` value. Fails against the previous query, passes with the fix.
- `TestDeleteElDataBeforeBlockUid` : deletes rows below the threshold and keeps the rest, under a deadline so a future re-introduction of the nested-transaction deadlock would be caught.

`go test ./db/`, `go vet`, and `gofmt` all clean.


pk910:
Heya @damilolaedwards
Thanks for the fixes, looking both good :)

pk910 (approved):

PR #779, opened 2026-06-30
fix: prevent out-of-range panic in validator and builder balance sort

damilolaedwards:
## Summary

The balance sort on the validators and builders pages can panic with an index out of range, which surfaces as a 500 on those pages.

`GetFilteredValidatorSet` and `GetFilteredBuilderSet` load balances from the epoch-start snapshot, but the result set is streamed from the canonical head. Validators projected from the pending-deposit queue, and builders onboarded mid-epoch, appear in the head set at indexes past that snapshot. The balance comparators indexed the balances slice directly, so sorting a set that includes one of those entries reads out of range and panics.

This is reachable in normal use: a non-empty pending-deposit queue is the common state after Electra, so the validators balance-descending sort is affected on mainnet today; the builders sort is affected once Gloas is live. No special input is needed, an ordinary request to the page with the balance sort triggers it.

The other accesses already guard the index (the ascending validator sort uses a bounds-safe helper, and the db-index sorts and result building check the length). Only the comparators were missed.

## Fix

Route the comparator lookups through a bounds-safe accessor that returns 0 for an index past the snapshot, so projected validators and freshly onboarded builders sort as zero balance instead of panicking. The validator descending branch now matches the ascending one, and the builders sort gets the same treatment as its sibling code paths.

## Tests

Added tests for the bounds-safe lookups and for sorting a set that contains an out-of-range index. They pass with the fix and fail with an index out of range without it.

`go build`, `go vet` and `gofmt` are clean.

### DIFF
--- a/services/chainservice_balance_sort_test.go
+++ b/services/chainservice_balance_sort_test.go
@@ -1,0 +1,68 @@
+package services
+
+import (
+	"sort"
+	"testing"
+
+	"github.com/ethpandaops/go-eth2-client/spec/gloas"
+	"github.com/ethpandaops/go-eth2-client/spec/phase0"
+)
+
+func TestBalanceAt(t *testing.T) {
+	balances := []phase0.Gwei{30, 10}
+	if got := balanceAt(balances, 0); got != 30 {
+		t.Errorf("in-range lookup = %d, want 30", got)
+	}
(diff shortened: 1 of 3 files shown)


pk910:
I've incorporated the fix into #783 as that PR reworks the builder handling and would conflict

PR #780, opened 2026-06-30
fix: handle contract-creation transactions in system-contract indexers

damilolaedwards:
## Summary

The system-contract event indexers (deposit, withdrawal, consolidation, builder deposit, builder exit) record a transaction target by dereferencing `tx.To()` directly in their event callbacks. Contract-creation transactions have no recipient, so `tx.To()` is nil for them.

The log filter matches by the system contract address with any topic, so a log emitted by a contract-creation transaction whose constructor calls a system contract is returned and routed into the callback, where the nil dereference panics. The panic is not caught by the per-log error handling (that only handles returned errors), and the indexer state advances only after the whole batch is persisted. The loop is restarted on panic and re-fetches the same block range, so it panics again on the same transaction and stops making progress, leaving every later event for that indexer unindexed.

This needs no special access. A transaction that calls a system contract from within a constructor (for example a contract that deposits during deployment) is enough.

## Fix

Resolve the recipient through a small shared helper that falls back to the emitting contract address when the transaction has no recipient, so creation-tx events are recorded against the system contract they reached instead of panicking. All ten call sites now use it.

## Tests

Added a test for the recipient lookup covering both a contract-creation transaction and a normal call transaction. It passes with the fix and panics without it.

`go build`, `go vet` and `gofmt` are clean.


indexer/execution/system_contracts/contract_indexer.go
```diff
@@ -173,6 +173,16 @@ func (ci *contractIndexer[_]) loadTransactionByHash(ctx context.Context, client
...
 }
 
+// txRecipient returns the transaction recipient. Contract-creation transactions
+// have no recipient (To is nil); for those the request reached the system contract
+// via an internal call, so the emitting contract address is the effective target.
+func txRecipient(tx *types.Transaction, log *types.Log) common.Address {
+	if to := tx.To(); to != nil {
+		return *to
+	}
+	return log.Address
```
pk910, line 183: nice fallback 👍👍

pk910 (approved):

PR #792, opened 2026-07-07
Add safe block visualization from fast_confirmation event stream

barnabasbusa:
## Summary
Screenshot 2026-07-07 at 14 42 13
image

Config to test with: 
```yaml
participants_matrix:
  el:
    - el_type: geth

  cl:
    - cl_type: lodestar
      cl_extra_params:
        - --chain.fastConfirmation=true
    - cl_type: nimbus
    - cl_type: grandine
      cl_image: ethpandaops/grandine:bomanaps-feature-fcr-minimal
      cl_extra_params:
        - --fast-confirmation-rule
    - cl_type: prysm
  count: 2
global_log_level: debug
network_params:
  preset: minimal

additional_services:
  - dora
```

Adds support for the `fast_confirmation` beacon event stream topic (fast confirmation rule, see [ethereum/beacon-APIs#616](https://github.com/ethereum/beacon-APIs/pull/616)) and visualizes the resulting **safe block** in the UI.

### Backend

- Subscribes to the `fast_confirmation` topic as an optional ancillary SSE stream per beacon node. Nodes that don't support the topic reject the subscription with a 4xx response, in which case the stream silently gives up — so the feature auto-activates only on FCR-enabled nodes, with no config needed.
- Parses the event payload (`block`, `slot`, and the optional `current_slot` added by beacon-APIs#616) with a local type, since the pinned go-eth2-client doesn't carry `current_slot` yet.
- Tracks the most recent fast confirmed (safe) block per client (`Client.GetLastFastConfirmation()`) and network-wide in the chain state (`ChainState.GetFastConfirmedBlock()`, monotonic by slot), plus a per-client `SubscribeFastConfirmationEvent()` dispatcher for future consumers.

### Frontend

- **Index page**: the network overview shows the safe slot next to the current slot (with a shield indicator), and canonical slots at/below the safe slot get a shield marker in the recent slots list. Both update live via the existing `/index/data` polling.
- **Consensus clients page**: new sortable "Safe Slot" column showing each node's last fast confirmed block (linked to the slot page). The column only appears when at least one node reports fast confirmations; nodes without FCR show a dash.
- **API**: `/v1/clients/consensus` now includes `fcr_enabled`, `safe_slot` and `safe_root` per client, and a new dedicated `/api/v1/network/fast_confirmation` endpoint returns the network-wide safe block (slot, root, epoch, last confirmation time) plus per-client fast confirmation status.

The UI additions are hidden entirely on networks where no node emits fast confirmation events.


clients/consensus/clientlogic.go
```diff
@@ -130,6 +130,10 @@ func (client *Client) buildEventStreamMask() uint16 {
 		events |= rpc.StreamInclusionListEvent
 	}
 
+	// fast confirmation is an optional node feature, the subscription is dropped
+	// silently if the node doesn't support the topic
+	events |= rpc.StreamFastConfirmationEvent
```
pk910, line 137: > the subscription is dropped silently if the node doesn't support the topic

Is this the case?  clients usually reject the whole subscription with a HTTP 400 for unknown topics.
We might need to use a separate stream

pk910 (approved):

PR #795, opened 2026-07-15
Validator name history resolver (inventory history snapshots)

qu0b:
## Problem

On devnets, validator names map index ranges to the node currently running them. When keys get reassigned, the mapping applies retroactively: historical views silently rewrite history, and only the on-chain graffiti still tells the truth.

## Scope (revised)

This PR is now **resolver infrastructure only** — no user-visible behavior change. A census of name resolution found 92 call sites across page handlers, API handlers and service-layer cache filters; converting them individually leaves the same event showing different names on different surfaces. The display/filter fix lands in a follow-up that stamps a `name_id` on event rows at write time (indexer-resolved, dictionary table), using this PR's resolver for live stamping and historically-correct backfill/resync.

## What this adds

The validator names inventory can serve an assignment history of full range snapshots (upstream: ethpandaops/ansible-collection-general#563):

```json
{
  "ranges": { "0-863": "prysm-besu-2" },
  "history": [
    { "effective_from": 0,          "ranges": { "0-863": "lighthouse-geth-1" } },
    { "effective_from": 1752308200, "ranges": { "0-863": "prysm-besu-2" } }
  ]
}
```

- `effective_from` is unix seconds (≤ genesis means "since genesis"); entries are full snapshots — idempotent, resync-safe
- Snapshots are converted to slot-bounded intervals via `ChainState.TimeToSlot` on the updater tick (retried until specs/genesis are known) and kept range-compressed in memory
- **`GetValidatorNameAt(index, slot)`**: resolves the name valid at a slot via two binary searches; any miss falls back to `GetValidatorName`, so inventories without `history` (and yaml/withdrawal-resolved names) behave byte-identically to today
- Malformed history entries (bad keys, overlapping or inverted ranges) are logged and skipped without touching the current-names path; range expansion is capped so a malformed range like `0-2^64` can't hang the loader (also applied to the existing `parseNamesMap`)

## Performance

Measured (100k validators, 10 snapshots × 100 ranges): `GetValidatorNameAt` history hit 9.1ns vs 17.4ns for the plain map lookup; miss+fallback 19.9ns; no-history networks 21.0ns. Zero allocations. Benchmarks committed.

## Testing

Unit tests for snapshot boundary conditions, fallback semantics, and history parsing (overlaps, bad keys, unsorted/shadowed entries). `go test ./...` green.


qu0b:
## Performance measurements (and a fix they forced)

Benchmarked both hot paths against the status quo before considering this mergeable.

**In-memory lookups** (100k named validators, 10 snapshots × 100 ranges, `-benchmem`):

| Path | ns/op | allocs |
|---|---|---|
| `GetValidatorName` (status quo) | 17.4 | 0 |
| `GetValidatorNameAt` history hit | 9.1 | 0 |
| `GetValidatorNameAt` miss → fallback | 19.9 | 0 |
| `GetValidatorNameAt` on no-history network | 21.0 | 0 |

History hits are *faster* than the map lookup; the worst case adds ~3.5ns/call. Irrelevant at page-render scale.

**Name-filtered slot queries** (200k slots, 100k validators all reassigned twice = 200k history rows, sqlite): the original range-containment join (`proposer BETWEEN range_start AND range_end`) could not use a btree — a full-scan no-match filter went **89ms → 5.0s (61×)**. Commit e414e31d restructures the table to per-index interval rows stored **only where the name differs from the current mapping** (COALESCE keeps semantics identical), making the join an index-seekable equality join:

| Case | legacy | history join |
|---|---|---|
| full-scan, no match (worst case) | 89ms | 161ms (1.8×) |
| selective match, LIMIT 25 | 27ms | 53ms (1.9×) |

This benchmark is deliberately pessimal (every validator reassigned within the window, sqlite); realistic devnets reassign a few percent and run postgres, landing much closer to legacy. Networks without history data still emit byte-identical SQL. Everything else (refresh cycle, DB writes, memory) is bounded by reassignment count — hundreds of rows, sub-millisecond, every few days.

Benchmarks are committed (`-bench`-only, no test-time cost) to guard against regressing this.

qu0b:
Scoped this PR down to the resolver infrastructure (see updated description). The read-side wiring (handler conversions, `validator_name_history` table, gated SQL join) is dropped: a call-site census found 92 name-resolution sites across pages, API v1 and cache filters, so per-site conversion can't be made consistent. Follow-up PR will stamp a `name_id` on event rows at write time (dictionary table + indexer resolution through `GetValidatorNameAt`), which fixes display and filtering on every surface at once and makes name filters an indexed integer equality.

qu0b:
Superseded: merged into #796 (the stacked branch contains these commits), which now targets master as the single PR for the whole feature — resolver infrastructure + name id stamping — so one image build covers it for devnet testing.

PR #796, opened 2026-07-15
Time-aware validator names for devnets (history resolver + name id stamping)

qu0b:
## Combined PR: time-aware validator names for devnets (supersedes #795)

On devnets, validator names map index ranges to the node currently running them. When keys get reassigned, the mapping applies retroactively: historical views silently rewrite history, contradicting on-chain graffiti. This PR fixes that end-to-end.

## Part 1 — history resolver (formerly #795)

The validator names inventory can serve an assignment history of full range snapshots (upstream: ethpandaops/ansible-collection-general#563):

```json
{
  "ranges": { "0-863": "prysm-besu-2" },
  "history": [
    { "effective_from": 0,          "ranges": { "0-863": "lighthouse-geth-1" } },
    { "effective_from": 1752308200, "ranges": { "0-863": "prysm-besu-2" } }
  ]
}
```

Snapshots are converted to slot-bounded intervals (`effective_from` unix seconds, ≤ genesis = since genesis) and exposed via `GetValidatorNameAt(index, slot)`, falling back to current names on any miss.

## Part 2 — name id stamping at write time

- `validator_names_dict(id, name)`: distinct names with stable ids
- `slots.proposer_name_id`: stamped by the indexer via an injected **cache-only** resolver (dict inserts happen on the updater tick only — a resolver-side insert deadlocks the finalization tx, found in e2e)
- `NULL` = unresolved (repaired at ~1.2M slots/hour, newest first), `0` = resolved-no-name
- **Reconciliation**: built snapshots are persisted in `explorer_state`; when history *resolution* changes, stamps reset from the first affected slot and get re-stamped — covers the inventory-refresh lag window, retroactive amendments, restarts
- **Gating**: stamping activates only on history-serving networks; yaml-based networks (sepolia/holesky/hoodi) keep byte-identical behavior (their names are renameable identity labels)
- Slots-family filters use indexed integer id matching with a legacy fallback arm for unstamped rows; display goes through `GetProposerName` with time-aware fallback

## Upgrade path (existing doras)

No resync needed: nullable-column migration, repair pass backfills automatically, deploy order vs the ansible role doesn't matter. Reassignments predating the ansible rollout collapse into the seed history entry.

## Validation

- Unit + sqlite integration tests (snapshot boundaries, filter semantics incl. invert/no-name/fallback arms, dict idempotency, repair roundtrip, diff-boundary logic)
- **E2E on a kurtosis devnet** (2×lighthouse/geth, minimal preset): write-time stamping, reassignment reconciliation at the exact computed boundary slot, boundary-sharp stamps, temporal filters with zero cross-era leakage, and a full DB-wipe resync reproducing historical stamps against a contradicting current mapping — details in comments below


qu0b:
Added b1fb2f1a after cross-checking against the inventory-history contract (ansible-collection-general#563): stamps are frozen at write time, but dora only refreshes the inventory every 2h — slots indexed between a key reassignment and the next refresh were stamped with the stale mapping, non-NULL, so the repair pass never revisited them.

Fix: the built snapshots are persisted in `explorer_state`; whenever a freshly built history *resolves differently*, stamps are reset to NULL from the first slot whose resolution changed and the repair pass re-stamps them. The diff deliberately ignores `endSlot` re-bounding, so a routine appended history entry only resets rows from the new boundary (~the refresh lag window, a few hundred rows), not from genesis. Retroactive history amendments and restarts are covered by the same mechanism; networks without history never write the state key.

Also hardened: a failed inventory load now keeps the previous snapshots and pauses stamping instead of clearing history — no stamp churn on transient outages.

qu0b:
## E2E validation on a kurtosis devnet

Ran the full feature against a local ethereum-package network (2× lighthouse/geth, minimal preset, 256 validators) with a local inventory endpoint serving the #563 history format, `validatorNamesRefreshInterval: 30s`.

**Caught and fixed a real deadlock (67938415):** the resolver runs inside the indexer's finalization write transaction; on a dictionary miss it opened a second write transaction → deadlock on sqlite's writer mutex (confirmed via pprof goroutine dump: `getOrCreateNameId → RunDBTransaction` blocked at `db/common.go:174`). The resolver is now cache-only — a miss stamps NULL and the repair pass heals it; dictionary inserts happen solely in `syncNameDict` on the updater tick.

**Verified end-to-end:**
1. Names + 1 history snapshot loaded from the inventory; dict synced; initial NULL backlog repaired in one batch (`stamped 8 slot rows`)
2. New slots arrive **stamped at write time** (repair ran exactly once)
3. **Reassignment**: moved validators 0-127 to `shuffled-node-a` via an appended history entry with `effective_from = now` — reconciliation fired at **exactly the expected boundary slot** (174, computed independently from genesis time), reset correctly scoped
4. **Boundary-sharp stamps**: slot 173 → `1-geth-lighthouse`, slot 174 → `shuffled-node-a`; 85 pre / 5 post rows, zero misstamps
5. **Temporal filters** via `/api/v1/slots?proposer_name=…`: old name → 147 slots all ≤ 173, new name → 125 slots all ≥ 174, zero cross-era leakage; display names historically correct per row
6. **Full resync** (DB wiped, re-indexed from CL): all 89 pre-boundary rows re-stamped with the *historical* name despite the current mapping saying otherwise; boundary intact; zero unstamped rows. Dict ids differed across runs — names resolve by content, id continuity not required.

This is the original bug fixed end-to-end: reassigning validators no longer rewrites history, and a resync reproduces it faithfully.

qu0b:
Added 9c8eb45b after thinking through the upgrade path for existing deployments: stamping (resolver, dict sync, repair) now activates **only on networks whose inventory serves an assignment history**. On yaml-based networks (sepolia/holesky/hoodi) names are identity labels that can be renamed retroactively — freezing them into rows would pin stale labels on historical slots. Those networks keep byte-identical behavior to master.

**Upgrade path for existing doras (no resync needed):**
1. Migration adds a nullable column + dict table (no table rewrite)
2. Devnets with inventory history: repair pass backfills stamps at ~1.2M slots/hour (10k per updater tick, newest first); display is correct immediately via the in-memory fallback, filters use the legacy arm until repaired
3. Devnets upgrading before the ansible role ships history: rows stay NULL until history appears, then reconciliation + repair backfill everything — deploy order doesn't matter
4. Reassignments predating the ansible rollout collapse into the seed history entry (the data doesn't exist retroactively)

qu0b:
@pk910 review feedback addressed:

**1. IN-list variable limits → dictionary join (7d099b3e).** You're right — expanding matched ids hits statement variable limits and costs a probe per id. The filter now does exactly what you suggested: `LEFT JOIN validator_names_dict ON validator_names_dict.id = slots.proposer_name_id` with the ILIKE on the joined name — a 1:1 replacement of the legacy join shape (one integer PK probe into a tiny cached table per row, no variable-count dependence). The in-memory id matching and `ProposerNameIds` plumbing are gone.

**2. Why the repair loop can't be a migration prefill.** A schema migration can only copy from `validator_names` = the *current* mapping — which is exactly the wrong value for slots before a reassignment boundary, and on first startup the reconciliation would detect the history mismatch and reset those rows anyway (prefilled ≠ NULL doesn't protect them). Correct values need runtime data a migration doesn't have: the inventory history *and* chain specs to convert `effective_from` timestamps to slots. The repair loop is that resolution, one-time per network, 10k rows/30s newest-first, and nothing is degraded while it runs — display resolves via the in-memory history fallback and the filter's legacy arm covers unstamped rows (your dict-join makes that explicit). After convergence it's a no-op (one indexed `IS NULL` probe per tick).

**3. Other queries joining validator_names — full census.** Nine more: voluntary_exits, slashings, withdrawals, withdrawal_requests(+txs), consolidation_requests(+txs), mev_blocks, and the validators-page filter. All intentionally keep current-name semantics in this PR — families migrate to stamped ids one at a time (each needs its own `name_id` column + writer change), slots first. And your included_deposits find was even better than a missing migration: the deposits name filter was **never implemented in SQL at all** — `ValidatorName` sits in the filter structs but `GetDepositsFiltered`/`GetDepositTxsFiltered` never consume it; only the in-memory unfinalized path applies it, which is why the page looks like it ignores the filter. Fixed independently of this PR in #799 (join through `validators.pubkey`, since deposits are pubkey-keyed).

PR #826, opened 2026-08-07
Drop named-fork entries from EL blob schedule (geth 1.17.5 compat)

barnabasbusa:
## Summary

go-ethereum v1.17.5 removed the named-hardfork fields (`Osaka`, `Amsterdam`, `UBT`) from `params.BlobScheduleConfig` (ethereum/go-ethereum#35029 "Remove named hardforks from the BPO schedule"). Named forks no longer carry their own blob schedule; they inherit from the BPO chain below them.

This PR removes the three rows referencing those fields from the fork table in `GetFullBlobSchedule` (`clients/execution/chainstate.go`), keeping Cancun, Prague, and BPO1-5. It still compiles against the currently-pinned v1.17.4 and unblocks the dora side of the dependabot bump in #814.

## Why this is behavior-safe

Both consumers of `GetFullBlobSchedule` are unaffected:

- `GetBpoForks` (`services/chainservice_forks.go`) only uses entries with `IsBpo == true`; the removed rows were all non-BPO.
- `GetBlobScheduleForTimestamp` picks the last entry before a timestamp. Under geth's new model a named fork never changes blob params — its schedule always equals the preceding BPO/Prague/Cancun entry — so those rows were redundant duplicates.

## Notes

The go-ethereum bump itself is not included here; it remains blocked on an ethcore update for geth 1.17.5, which is in progress separately.

### DIFF
--- a/clients/execution/chainstate.go
+++ b/clients/execution/chainstate.go
@@ -242,6 +242,8 @@ func (cache *ChainState) GetFullBlobSchedule() []BlobScheduleEntry {
 	chainConfig := cache.config.Config
 	blobConfig := chainConfig.BlobScheduleConfig
 
+	// Named forks (Osaka/Amsterdam/UBT) no longer carry their own blob schedule
+	// since geth 1.17.5 and inherit from the BPO chain below them.
 	forks := []struct {
 		time      *uint64
 		schedule  *params.BlobConfig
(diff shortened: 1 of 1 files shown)


clients/execution/chainstate.go
```diff
@@ -242,21 +242,20 @@ func (cache *ChainState) GetFullBlobSchedule() []BlobScheduleEntry {
 	chainConfig := cache.config.Config
 	blobConfig := chainConfig.BlobScheduleConfig
 
+	// Named forks (Osaka/Amsterdam/UBT) no longer carry their own blob schedule
+	// since geth 1.17.5 and inherit from the BPO chain below them.
```
barnabasbusa:
```suggestion
```

qu0b (approved):

barnabasbusa:
Closing in favor of #814, which now carries this change (the blob-schedule fix branch was merged into the dependabot branch there).

PR #832, opened 2026-08-14
feat: ENS name search in the main search bar

barnabasbusa:
## Summary

Adds **forward ENS resolution** (`vitalik.eth` → address) to the main search bar, building on the existing reverse-resolution (address → name) subsystem — no new dependencies.

- Typing a complete ENS name shows an **"ENS Names"** typeahead section with the resolved address (plus a `Contract` badge when applicable); selecting it goes to `/address/0x…`
- Submitting the search form with an ENS name redirects straight to the address page (checked after the numeric/hex resolvers, before validator-name/graffiti fallbacks; falls through if the name doesn't resolve)
- Subdomains (`sub.name.eth`), non-`.eth` TLDs and emoji names are supported; a shape check (complete name, TLD label ≥ 3 chars) avoids per-keystroke eth_calls while typing

## ENS is always resolved against mainnet

ENS lives on Ethereum mainnet, so lookups now always use the dedicated `ensResolver.endpoints` — **defaulting to `https://ethereum-rpc.publicnode.com` when unset**. The previous fallback to the local execution pool is removed: on devnets/testnets that pool serves a chain without an ENS deployment, so it could never resolve anything. This applies to the existing reverse resolution too. Operators who don't want the public-RPC dependency can point `ensResolver.endpoints` at their own mainnet node.

## Implementation notes

- `services/ensresolver_ens.go`: new `resolveForward()` — `registry.resolver(namehash(name))` → `resolver.addr(node)`, reusing the existing namehash/calldata/eth_call helpers and registry priority order
- `services/ensresolver.go`: exported `ResolveEnsName()` with its own LRU cache (positive/negative entries honor `refreshPositive`/`refreshNegative`), 10s call timeout
- `handlers/search.go`: ENS branch in the full-search resolver + new `ens` search-ahead type; the global `0x` strip is skipped for ENS queries since labels like `0xdeadbeef.eth` are legal
- Feature is gated on `ensResolver.enabled && executionIndexer.enabled` (results redirect to `/address`, which requires the indexer)
- ENS names in typeahead suggestions are HTML-escaped (same as graffiti results)

Same known limitations as the existing reverse path: lowercase normalization only (no UTS-46/ENSIP-15), no wildcard/CCIP-read (ENSIP-10) resolvers.

## Testing

- `go build ./...` and `go vet` clean
- Name-shape regex verified against `vitalik.eth`, `sub.vitalik.eth`, `0xdeadbeef.eth`, `nick.xyz`, emoji names, and partial-keystroke negatives

### DIFF
--- a/config/default.config.yml
+++ b/config/default.config.yml
@@ -192,8 +192,9 @@ rpcProxy:
 # Optional ENS resolver: resolves execution addresses to their primary ENS name.
 ensResolver:
   enabled: false # enable ENS name resolution & display
-  # optional dedicated EL RPC endpoints (usually Ethereum mainnet). If empty, an
-  # available client from the main execution pool is used.
+  # EL RPC endpoints used for ENS lookups. ENS lives on Ethereum mainnet, so these
+  # should always point at a mainnet RPC regardless of the chain being explored.
+  # If empty, a public mainnet RPC (ethereum-rpc.publicnode.com) is used.
   endpoints: []
   #  - url: "https://ethereum-rpc.publicnode.com"
   #    name: "mainnet"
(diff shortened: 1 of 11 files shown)


barnabasbusa:
Closing in favor of #836

PR #863, opened 2026-09-03
Speed up page builds on large validator sets and huge block access lists

pk910:
Two sets of page-build fixes found by profiling production devnet instances: the validator pages on glamsterdam-devnet-9, and the slot / transaction pages of blocks with huge access lists on glamsterdam-devnet-8.

# Validator pages on large validator sets

## Problem

`/validators`, `/validators/activity` and `/validators/summary` on glamsterdam-devnet-9 take 5-17 s per uncached build. The pod is limited to 1.5 CPU cores and the indexer's state replay (hashtree via cgo) occupies most of that, so anything CPU-heavy in a page build stretches badly. The network has 4M validators (3M exited, 1M active), so every page build streams 4M cache entries.

CPU profiles (`/debug/pprof/profile`, ≤14 s because of the write timeout) and goroutine dumps on the production instance showed:

- **activity page** (4.6 s CPU per build): 2.8 s in `GetValidatorName` and 1.4 s in `GetValidatorLiveness`, both a lock plus a random lookup in a million-entry map per validator.
- **validators list** (5.4 s CPU per build): 98 % in `db.GetValidatorIndexesByFilter`, which fetched all 1M matching indexes from Postgres through sqlx on every page just to page through them.
- **summary page**: the same per-validator lookups plus the full filtered-set merge with DB query, and client type parsing per validator (1 GB of cumulative allocations).

## Changes

- **Validator names as ranges**: inventory ranges were expanded into a per-index map. Names are now kept as sorted, disjoint ranges resolved by binary search from an immutable index published through an atomic pointer, so hot-path lookups are lock-free (5 ns in the benchmark). Overlapping ranges keep the previous precedence (later loaded source wins). Resolved names (withdrawal/depositor based) stay a map layered underneath. The db sync expands ranges when writing `validator_names`.
- **Bulk liveness / inclusion accessors**: `GetValidatorLivenessCounts` and `GetValidatorInclusionDistances` walk the activity cache once and return slices indexed by validator index. Used by the activity, offline and summary pages and the activity API.
- **Bounded DB fetch in `GetFilteredValidatorSet`**: the index query gets a `LIMIT` of the requested window plus the number of cached entries that have a DB row (cached matches only take positions away from DB rows, and only entries within the persisted range can supersede one), which holds whenever the db orders by the same key the merge sorts by. The balance orders merge by the current balance from the beacon cache while the db can only order by effective balance (which most validators tie on), so those two keep the full fetch. A cached state is newer than its db row (the cache only carries a validator object until its state is persisted), so cached indexes supersede their db rows in the merge even when the newer state no longer matches the filter. The total is computed to equal what the full merge would list: db `COUNT` minus the matching rows superseded by a cached state (counted with the same filter predicate, in chunks) plus every cached match. While the superseded backlog is large (restart against a stale db, a big finalization update still persisting) the full fetch is used, whose merge counts exactly. Cached-index membership uses a bitset instead of a map. Callers without a limit (name resolver, exact withdrawal search) keep the full fetch.
- **Summary streams the cache**: active validators come straight from `StreamActiveValidatorData` (index, effective balance, slashed flag are all in the cache), client type parsing is memoized per name. Projected pending-deposit placeholders (far-future eligibility epoch) are skipped, as the status resolution used before excluded them.
- **Stable paging**: every sort order breaks ties by validator index, in the query and in the merge. A bounded top-N sort returns tied rows in an order that changes with the limit, and the epoch and effective-balance keys tie for most validators on a devnet.
- **Activity page** formats index-group labels once per group and lowercases names once per name.

## Measurements

Uncached builds, local instance against the devnet-9 endpoints (same 4M validators; the local validators table was still being filled, so the DB fetch itself is not represented), before (production) → after (local):

| page | before | after |
|---|---|---|
| /validators/activity | 5-14 s | 0.25 s |
| /validators/summary | ~10 s CPU | 0.5 s |
| /validators (all sort orders and filters tried) | 7-17 s | 0.2-0.6 s |

On production the validators list additionally fetches `offset + limit + (validators changed since restart)` index rows instead of all 1M. Summary numbers match production for the same epoch (1,000,053 validators, effective balance within a few ETH from timing).

## Notes

- Verified the bounded total against the exact full walk on a local devnet-8 instance for every sort order and filter, including rows edited in the db to disagree with the cache.
- Deep pages far into the persisted range still stream DB rows up to the offset, as before.
- Sorting by balance is still the slow case on large sets: it needs the full matching index set, since only the beacon cache knows current balances.

# Slot pages with huge block access lists, and block loads outside the cache

## Problem

Slot 176225 on glamsterdam-devnet-8 (EL block 168569, 11 transactions, 184M gas) touches 58,387 addresses, 58,366 of them without any state change. The access list tab rendered one collapsible card per address server-side and embedded the whole decoded list a second time as JSON for the client-side detail renderer, so the page was 72 MB of HTML (raw access list RLP: 1.5 MB), re-rendered on every request since only the page model is cached. Production logs show `write: i/o timeout` and `connection reset by peer` on that route: clients gave up before the response finished.

Uncached transaction pages of the same block took 4-10 s regardless of the transaction. Goroutine dumps showed every in-flight build in `loadFullTransactionData → GetSlotDetailsByBlockroot`, fetching the beacon block header, body and Gloas payload envelope from the beacon nodes. For a block outside the indexer cache the chain service asked the nodes first and used blockdb only as a fallback, and on devnets the nodes are overloaded (context deadline exceeded across the EL and CL pools in the logs). The payload envelope carries the access list, so it is 1.58 MB for this block against a 4 KB block body.

## Changes

- **Lazy access list**: `/slot/{slotOrHash}/bal` (`handlers/slot_bal.go`) returns the decoded entries as JSON (accepts a slot number or block root like the slot route, cached 30 min). The path value is resolved against the indexed blocks first and unknown blocks answer 404, so only known blocks reach the block loader. The page model keeps only `BALSummary`; the tab badge, download entry and tab visibility key off it. The tab fetches the entries when opened and renders paged cards client-side with an address filter, a "hide addresses without changes" toggle and a page-size selector. Card details render on expand as before, the compare modal waits for the list to load.
- **Links inside the cards**: the address in a card header links to the address page and "Tx N" badges link to the transaction page (same resolution as the server-side link helpers, via the new `ethAddressLinkBase` / `ethTransactionLinkBase` template funcs). Bootstrap's collapse data-api runs in the capture phase and cancels clicks on links inside a toggle, so the header now toggles its collapse from its own click handler that ignores links.
- **Node-vs-blockdb race** (`raceBlockLoad` in `services/chainservice_blocks.go`, used by both the by-root and by-slot loaders): the beacon nodes are still asked first. After `blockDb.parallelLoadDelay` (default 5s), or as soon as the nodes gave up, the block db is asked for the same block in parallel and whichever source returns a block first wins; the loser is cancelled. `0s` starts both loads at once, an explicit `null` or a negative value keeps the previous sequential order. Node loading itself is unchanged, only moved into `loadBlockFromClientsByRoot` / `loadBlockFromClientsBySlot`. Table-driven test with the race detector in `chainservice_blocks_test.go`.

## Measurements

| Slot 176225 | before | after |
|---|---|---|
| HTML served | 72 MB | 273 KB |
| access list entries | inline | 2.5 MB JSON from `/slot/176225/bal`, 0.3 s cached |

Verified on a local instance against devnet-8 with headless Chrome driven over CDP: 100 of 58,387 entries per page, filter down to the 21 changed entries, expanded card content, "Tx N" badge linking to its transaction page, header address link navigating to the address page. The blockdb-wins path could only be exercised through the unit test; on the local run the nodes always answered first.


db/validators.go
```diff
@@ -201,6 +224,11 @@ func GetValidatorIndexesByFilter(ctx context.Context, filter dbtypes.ValidatorFi
 		fmt.Fprintf(&sql, " ORDER BY withdrawable_epoch DESC")
 	}
 
+	if limit > 0 {
+		args = append(args, limit)
+		fmt.Fprintf(&sql, " LIMIT $%v", len(args))
```
qu0b, line 231: With a `LIMIT` the `ORDER BY` needs a tiebreaker. `activation_epoch`, `exit_epoch`, `withdrawable_epoch` and `effective_balance` (used when `balances == nil`) have huge tie groups on devnets (every genesis validator shares activation epoch 0, nearly all share `FAR_FUTURE` exit/withdrawable epoch), and the bounded fetch in `GetFilteredValidatorSet` runs this query with a different `LIMIT` per page (`offset + limit + superseded`). Postgres' top-N sort orders tied rows differently for different N, so page 2 can repeat rows from page 1 and skip others.

Reproduced on Postgres 15 with 1M rows tied on `activation_epoch`: the first 50 rows of `ORDER BY activation_epoch LIMIT 100` are not the 50 rows returned by `LIMIT 50`, and a page-1 row shows up again in positions 51-100. Before this PR the single unbounded sort kept the pages consistent (there is no index on these columns, so the plan is seq scan + top-N heapsort).

Fix: append `validator_index` as a secondary key to every `ORDER BY` in this function, and tie-break on `Index` in the `sortFn` cases of `GetFilteredValidatorSet` so the merge and the db share one total order.

templates/slot/block_access_list.html
```diff
@@ -811,50 +785,319 @@ <h6 class="mb-3">
...
+        });
+      }
+    };
+
+    // balTxHref returns the link target of a block access index that maps to one of
+    // the block's transactions (index 0 and txCount+1 are the pre/post execution
+    // system operations, which have no transaction page).
+    function balTxHref(txIndex) {
+      if (!window.balTxLinkBase || typeof transactionDetails === 'undefined') return '';
+      if (txIndex < 1 || txIndex > transactionDetails.length) return '';
```
qu0b, line 1023: `transactionDetails` is `null` for blocks without transactions: `handlers/slot.go` only fills `TransactionDetails` when the payload has transactions, so `JSON.parse({{ includeJSON .TransactionDetails false }})` yields `null`. `typeof null` is `'object'`, so the guard on the line above passes and `transactionDetails.length` throws a `TypeError`.

An empty block still has post-execution BAL entries (withdrawals / fee recipient at block access index txCount+1 = 1), so with tx links enabled (execution indexer or `ethExplorerLink` set) expanding such a card throws inside `renderTabContent` before `innerHTML` is set and the card body stays blank.

Fix: `if (!window.balTxLinkBase || !Array.isArray(transactionDetails)) return '';`

handlers/validators_summary.go
```diff
@@ -192,13 +191,22 @@ func buildValidatorsSummaryPageData(ctx context.Context) (*models.ValidatorsSumm
...
 	activeValidators := uint64(0)
 
-	for _, validator := range validators {
-		if validator.Validator == nil {
-			continue
+	// The active set is streamed straight from the validator cache: the summary only needs
+	// index, effective balance and the active/slashed state, which the cache carries for
+	// every validator without loading the full validator objects.
+	services.GlobalBeaconService.StreamActiveValidatorData(false, func(validatorIndex phase0.ValidatorIndex, validatorFlags uint16, activeData *beacon.ValidatorData, _ *phase0.Validator) error {
+		if activeData == nil || activeData.ActivationEpoch > currentEpoch || activeData.ExitEpoch <= currentEpoch {
```
qu0b: This predicate counts projected pending-deposit placeholders as active validators. `streamValidatorSetForRoot` streams them with `activeOnly=false`, and `validatorFromDeposit` gives them `ActivationEligibilityEpoch = FarFuture`, `ActivationEpoch = estimated processing epoch`, `ExitEpoch = FarFuture`, `EffectiveBalance = 0`. The estimate starts at `stateEpoch + 1`, and the state feeding the projection is the dependent-root state of the previous epoch, so the first batch of placeholders has `ActivationEpoch == currentEpoch` for the whole epoch. They pass this check and are counted as active, unslashed, zero-liveness validators in whichever client cell their index-range name maps to, inflating `ValidatorCount` and `OfflineValidators`.

The previous path went through `ValidatorToState`, which returns pending for `ActivationEligibilityEpoch == FarFuture` and excluded them.

Fix: also skip `activeData.ActivationEligibilityEpoch == beacon.FarFutureEpoch` here (or reuse the `ValidatorToState` predicate).

templates/slot/block_access_list.html
```diff
@@ -190,72 +190,46 @@
...
+      <div id="balLoading" class="text-center py-4">
+        <div class="spinner-border text-primary" role="status">
+          <span class="visually-hidden">Loading...</span>
         </div>
-      {{ end }}
+        <div class="mt-2 text-muted">Loading access list...</div>
+      </div>
+      <div id="balError" class="alert alert-danger" style="display: none;"></div>
+      <div id="balEntries"></div>
+      <div id="balPager" class="d-flex justify-content-between align-items-center flex-wrap gap-2 mx-2 mt-2" style="display: none;">
```
qu0b: `d-flex` and the inline `display: none` conflict: the bundled Bootstrap declares `.d-flex{display:flex!important}`, which beats both the inline style and the `pager.style.display = 'flex' | 'none'` toggle in `renderPage`. So the pager row (empty info text, disabled buttons) is visible while the list loads and stays visible for lists that fit on one page.

Fix: drop `d-flex` and the inline style from the markup and toggle the classes instead, e.g. `pager.classList.toggle('d-flex', visible); pager.classList.toggle('d-none', !visible);` with `d-none` set initially.

qu0b (commented):
Reviewed the full diff, built it and ran the changed packages with `-race` (green). Four inline comments below on things I'd fix before merging; each was verified against the current head.

config/default.config.yml
```diff
@@ -149,6 +149,11 @@ database:
 blockDb:
   engine: "none" # pebble / s3 / tiered / none (disable block db)
 
+  # blocks that are not in the indexer cache are fetched from the beacon nodes first;
+  # after this delay the block db is asked in parallel and the first answer wins.
+  # 0 = start both loads at once, null = sequential (nodes first, block db as fallback)
```
qu0b: The documented `0` value does not parse. `ParallelLoadDelay` is a `*time.Duration`, and yaml.v3 (v3.0.1 from go.mod) deliberately refuses an int scalar for `time.Duration`, so `parallelLoadDelay: 0` makes `ReadConfig` fail and `main` exits via `logrus.Fatalf("error reading config file: ...")`.

Reproduced against the same yaml.v3 version with the `blockDb` struct:

```
parallelLoadDelay: 0     -> yaml: unmarshal errors: line 2: cannot unmarshal !!int `0` into time.Duration
parallelLoadDelay: 0s    -> 0s
parallelLoadDelay: null  -> nil
parallelLoadDelay: 5s    -> 5s
```

`types/config.go:202` makes the same claim ("0 starts both loads at once"). Fix: document `0s` here and in the field comment (the env var is unaffected, `time.ParseDuration("0")` is valid).

qu0b (approved):

PR #873, opened 2026-09-21
signal in- vs out-of-protocol builder bids on the proposer build-source icon

qu0b:
### Description

On Gloas+ blocks, the proposer column already shows a build-source icon: a house for self-built payloads and a hard-hat linking to the builder for builder-built payloads. But there is no way to tell *how the winning bid reached the proposer* — via the in-protocol p2p gossip path (`SignedExecutionPayloadBid`) or out-of-protocol via the builder API / a relay.

This PR extends the hard-hat icon with that signal, using the bid gossip observation counters that dora already tracks (`SeenCount` / `SeenTotal` from the bid cache and `block_bids` table, added in 7dab4b26):

| Observation | Meaning | Rendering |
|---|---|---|
| `SeenTotal > 0 && SeenCount > 0` | **In-protocol** bid (observed on gossip) | default helmet, tooltip: `in-protocol bid (N/M gossip)` |
| `SeenTotal > 0 && SeenCount == 0` | **Out-of-protocol** bid (builder API / relay only) | amber helmet (`.build-source-oop`), tooltip: `out-of-protocol bid (no gossip observation)` |
| `SeenTotal == 0` | no observation data | default helmet |

Notes:
- Counters are derived at page-render time from the bid cache plus `block_bids` (new `db.GetBidsForSlots` slot-set query + `maxSlot` bound on `GetBidsForSlotRange`), so no schema change is needed and recently cached (not yet flushed) bids are covered.
- Wired on all pages that render the build-source icon: slots, blocks, filtered slots, and the index page (both the Go template and the knockout JS mirror in `page-index.js`).
- Includes a template-execution regression test for the new branches (`utils/format_buildsource_test.go`).


qu0b:
Closing per @pk910's feedback — verified against devnet-8 data and he's right: gossip observations can't classify the bid delivery path on Gloas.

Data from the cited example ([slot 247251](https://dora.glamsterdam-devnet-8.ethpandaops.io/slot/247251#bids), via `/slot/247251/bidseen`):
- The **winning bid** (builder 1889, `api-buildoor-teku-nethermind-1`, builder-API path) was seen on gossip by **29/94 clients** starting −343 ms before slot start, first by `lodestar-besu-1` — API-delivered bids get re-published on the `execution_payload_bid` topic.
- Several bids share the **same payload hash** (`0x1aabdb87…`) under 4 different builder indices, first observed by `buildoor-*` nodes with 26–53/94 reach — builders double-publish via API *and* gossip.
- Some bids were observed by exactly 1 client (the buildoor's own node), showing the SSE bid event also fires for API-received bids.

This matches the [Gloas p2p spec](https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/p2p-interface.md): the `execution_payload_bid` topic is designed to propagate bids network-wide regardless of how the builder first delivered them, so "seen on gossip" measures bid reach, not delivery path. As implemented, the icon would mislabel out-of-protocol bids as in-protocol.

If out-of-protocol categorization is still wanted, a more reliable signal would be relay-side delivery data (relay data APIs / xatu), similar to how mev-boost relay attribution works today.

